PR #264, opened 2012-11-01
Introducing Casper.blockStep

starlocke:
There exists circumstances where the next CasperJS step executes too early, causing false positives for failures in subsequent testing.

This commit introduces `Casper.blockStep()`, also integrated into `click()` and `clickLabel()`. Developers will now be able to give CasperJS a hint about the nature of a click, and expect CasperJS to block execution of its next step until some not-yet-started behaviour actually starts and completes (today, we're concerned about `loadInProgress` and `navigtationRequested` -- `pendingWait` exists, but I can't think of why I'd want to invoke `blockStep` using that reason).

It is relatively easy to maintain if new reasons for blocking the next step are created in the future; just add a new `case` for `switch` within `blockStep()`. It is also possible to directly call `blockStep()`, but a developer would typically be better just directly setting `Casper.navigationRequested`, or `Casper.loadInProgress`, or `Casper.pendingWait`.

The idea of introducing blockStep() tied into click() and clickLabel() is so that we express the whole thing as a single statement.

This usage is probably the generalized version:

```
Casper.click("#that_button", "loadInProgress");
```

This usage is probably more specifically looking at navigation of the `_top` frame (`isMainFrame` in CasperJS code):

```
Casper.click("#that_button", "navigationRequested");
```

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -228,7 +228,7 @@ Casper.prototype.base64encode = function base64encode(url, method, data) {
  *
  * @param   String|Array   reasons   One or more of: "loadInProgress, navigationRequested, pendingWait"
  */
-Casper.prototype.blockStep = function blockStep(reasons) {
+Casper.prototype.holdOn = function holdOn(reasons) {
     "use strict";
     if(utils.isString(reasons)){
         reasons = [reasons];
@@ -241,6 +241,7 @@ Casper.prototype.blockStep = function blockStep(reasons) {
             switch(blockParam){
                 case 'navigationRequested':
                 case 'loadInProgress':
+                case 'pendingWait':
                     this[blockParam] = true;
                 break;
                 default:
@@ -407,14 +408,34 @@ Casper.prototype.clear = function clear() {
  *
  * In case of success, `true` is returned, `false` otherwise.
  *
- * @param  String         selector           A DOM CSS3 compatible selector
- * @param  String|Array   blockStepReasons   One or more of: "loadInProgress, navigationRequested, pendingWait"; see Casper.blockStep
+ * @param  String   selector   A DOM CSS3 compatible selector
+ * @param  Object   options    Click options
  * @return Boolean
  */
-Casper.prototype.click = function click(selector, blockStepReasons) {
+Casper.prototype.click = function click(selector, options) {
     "use strict";
     this.checkStarted();
-    this.blockStep(blockStepReasons);
+    options = utils.isObject(options) ? options : {};
+    for(var option in options) {
+      var values = options[option];
+      switch(option){
+        case 'holdOn':
+          var funcCall = 'this.' + option + "(values);";
+          eval(funcCall);
+        break;
+        case 'loadInProgress':
+        case 'navigationRequested':
+        case 'pendingWait':
+          if(!utils.isType(values, 'boolean')){
+            throw new CasperError(f("Click option `%s` must be either true or false", option));
+          }
+          this[option] = values;
+        break;
+        default:
+          throw new CasperError(f("Casper does not recognize `%s` as a click option", option));
+        break;
+      }
+    }
     return this.mouseEvent('click', selector);
 };
 
@@ -422,18 +443,18 @@ Casper.prototype.click = function click(selector, blockStepReasons) {
  * Emulates a click on the element having `label` as innerText. The first
  * element matching this label will be selected, so use with caution.
  *
- * @param  String       label              Element innerText value
- * @param  String       tag                An element tag name (eg. `a` or `button`) (optional)
- * @param  String|Array blockStepReasons   One or more of: "loadInProgress, navigationRequested, pendingWait"; see Casper.blockStep
+ * @param  String   label     Element innerText value
+ * @param  String   tag       An element tag name (eg. `a` or `button`) (optional)
+ * @param  Object   options   Click options
  * @return Boolean
  */
-Casper.prototype.clickLabel = function clickLabel(label, tag, blockStepReasons) {
+Casper.prototype.clickLabel = function clickLabel(label, tag, options) {
     "use strict";
     this.checkStarted();
     tag = tag || "*";
     var escapedLabel = label.toString().replace(/"/g, '\\"');
     var selector = selectXPath(f('//%s[text()="%s"]', tag, escapedLabel));
-    return this.click(selector, blockStepReasons);
+    return this.click(selector, options);
 };
 
 /**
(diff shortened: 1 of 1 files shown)


brettjonesdev:
Hey folks, so is this still just in limbo?  I've run into these issues and spent a lot of time pulling my hair out over the non-deterministic failures I bump into when testing an Ajax-heavy application.  

I have for a long time littered my tests with `casper.wait()` calls, opting more recently for `casper.waitUntilVisible` where possible, but even now I still get random errors.

Is the hangup here just that we can't decide on a style for the feature @starlocke has proposed?  I'm a big fan of the tool and hope to use it for a long time to come, but having random failures like I am right now is a real productivity killer and prevents me from integrating my test suite into our CI setup.  

Is there at least a definitive workaround for now?

starlocke:
Hello @brettjonesdev this is indeed in limbo right now owing to my lack of time working with CasperJS. You mentioned "littering your code with wait()"? Have you tried littering your code with my proposed workaround, though -- namely... "[...] a developer would typically be better just directly setting Casper.navigationRequested, or Casper.loadInProgress, or Casper.pendingWait."?

Typical use case: Set one of those flags to "true" before doing a click(), or clickLabel(), or any other navigation or AJAX-ish thing - pick the one whose name sounds the most like what you're doing. You can dig into the Casper source code to choose the most appropriate one.

I used to try using wait(), but that is not at all the correct workaround due to the real nature of the "then()" function as explained in the StackOverflow Q&A -- thus inducing me to dig deep, figure it out, and finally to produce this pull request.

starlocke:
This pull request is simply a convenience that improves the API.  The baseline workaround can still be invoked without this pull request ever going into production -- just set "true" to one of the appropriate CasperJS flags: **Casper.navigationRequested**, or **Casper.loadInProgress**, or **Casper.pendingWait**

brettjonesdev:
Thanks a lot, @starlocke!   I'll give that a shot for sure.  It would be nice if the Casper docs mentioned this issue somewhere, explaining the need to tell Casper that a click triggers asynchronous events and explaining how flags ought to be used.  I couldn't find documentation for those flags anywhere in the docs.  

I'll give your suggestion with the flags a shot, but if this *really is* the only way to avoid this behavior, that's something that really ought to be documented at the least.  Eh, @n1k0 ?

n1k0:
> that's something that really ought to be documented at the least

That's a good idea. Would you mind creating a docs PR for this?

brettjonesdev:
I'd be happy to help document this, but I'm pretty ignorant of how these flags work right now, so let me make sure I understand how to use them if you don't mind.  

Let's say I have something like the following, a link click which triggers an Ajax call (say) and then causes an element to show:

```javascript
casper.click("#loadModal");
// casper.loadInProgress = true ???
casper.waitUntilVisible("#modal");
casper.then(function() {
   console.log("Modal visible");
})
```

Does it matter which of the 3 flags I set?  Also, if I set one of those flags after a click which does NOT result in asynchronous DOM changes in the browser, will that mess things up?  I'm thinking of writing a wrapper for `click()` in my utils class which will always just set one of those flags, wondering what the consequences of that approach would be.

starlocke:
@brettjonesdev - A minimal change is to simply raise the appropriate flag(s) before your calling `click()`. My test case code first raised `loadInProgress` alone, then it raised that plus `navigationRequest`. One or the other is sufficient, if memory serves me right. If you're going to do a JavaScript asynchronous effect, then, maybe you'll be raising the flag for `pendingWait` just before calling some kind of `wait()` method.

I added "FLAG" as a comment to mark manually raised flags. Raising flags is exactly what this pull request fulfills as a utility function.

```JavaScript
var casper = require('casper').create({
  viewportSize: {width: 1280, height: 768}
  //, verbose: true
  , logLevel: 'debug'
});
casper.start('http://www.example.com/sleepy_page.html');
casper.then(function (){
  this.test.info('home');
  this.test.info(this.getCurrentUrl() + '\n');
  this.fill("form", {
    'hello' : 'world'
  }, false);
  this.loadInProgress = true; // FLAG ONE -- equivalent to doing Casper.blockStep(), or using an augmented click() [neither of those yet implemented officially]
  this.click('#my_form_submit');
});
casper.then(function (){
  this.test.info('should be hello=world');
  this.test.info(this.getCurrentUrl() + '\n');
  this.fill("form", {
    'hello' : 'foobar'
  }, false);
  this.loadInProgress = true; // FLAG ONE -- equivalent to doing Casper.blockStep(), or using an augmented click() [neither of those yet implemented officially]
  this.navigationRequest = true; // FLAG ANOTHER -- equivalent to doing Casper.blockStep(), or using an augmented click() [neither of those yet implemented officially]
  this.click('#my_form_submit');
});
casper.then(function (){
  this.test.info('should be hello=foobar');
  this.test.info(this.getCurrentUrl() + '\n');
});
casper.run();
```

brettjonesdev:
Thanks, guys.  Interestingly, I haven't seen this kind of issue crop up again in several days and dozens of test runs... Once it starts to bite me again I'll try your suggestion with raising flags.  

The only question I had about that, though, was if I set `loadInProgress` (say), won't that flag stay set indefinitely, potentially causing problems?  I.e. wouldn't I need to set it to false in my `.then()` callback?  

I.e. something like this:

```
casper.loadInProgress = true;
casper.click('#button');
casper.then(function() {
   casper.loadInProgress = false;
   //do something
});
```

Does that make sense?  I'm just wondering if that flag would stay set indefinitely, which might not be a good thing, eh?

thiswillbeyourusername:
I wrote a script before reading this pull request, well hopefully my solution will help you.

Rather than wait for progress/page change etc, I use the [MutationObserver](https://developer.mozilla.org/en/docs/Web/API/MutationObserver) API. The code waits for *something* to happen in the DOM (within a selector) and then fires the event.

An example of it in action:
```
this.MutationObserver({
    selector: 'form.login-page',
    exclude: ['#loading-bar']
});

this.click('input[type=submit]');

this.waitForMutation(function() {
    // ...
});
```

It will wait for (1) dynamic JS onsubmit errors list being added into form.login-page, or (2) AJAX loading of login and then changing the page dynamically, or (3) a normal redirection.

`exclude` will ignore DOM changes to certain elements within the selector, such as a loading bar.

If you're interested in the code, I can post it but it's using jQuery so it would need re-writing for Casperjs.

BIGjuevos:
Do we want to keep this issue in the 1.1 release @istr @mickaelandrieu? I vote to move to a new milestone 1.2.

I feel like this would be good for a 1.2 considering our focus on squashing bugs for 1.1.

mickaelandrieu:
Agree

kensoh:

PR #384, opened 2013-02-16
Added git bash command

jfparadis:
I've added a shell script to use CasperJS with Unix on Windows (i.e. Git Bash):

$ casperjs test.js

CasperJS already ships with a Windows batch script to use it without Python nor Ruby. However, that script only works from within Windows command line (or from utilities such as grunt).

### DIFF
--- a/batchbin/casperjs
+++ b/batchbin/casperjs
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+export CASPER_PATH=$(dirname $0)/..
+export CASPER_BIN=$CASPER_PATH/bin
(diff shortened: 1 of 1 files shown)


paazmaya:
As @mduvall mentions, please place the script under `bin` directory, since `batchbin` no longer exists, and name it `casperjs.sh`.

Any documentation additions needed, perhaps telling briefly about this script and stating that it is PhantomJS specific?

istr:
Like @hexid pointed out there is sufficient support by means of the mono builds. Closing.

PR #486, opened 2013-05-22
allow users to indicate whether need override MimeType when performing ajax Requests

RockLi:
Current the mimeType is fixed in the codes and already overrided,
this will block some ajaX requests which return back those like Chinese in some scenarios..


kensoh:

PR #518, opened 2013-07-04
Allow to get a property of a remote object

clochix:
This allow to get a property of a remote object.

Usage: 

```js
Casper.getGlobal('foo[0].bar.baz');
```

(in my use case, I can not use `Casper.getGlobal('foo')[0].bar.baz` because foo contains circular references and `JSON.stringify` fails).

I presume this patch is not a very good idea, too complex, and I should just use evaluate to get the property I need.

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -1081,8 +1081,16 @@ Casper.prototype.getGlobal = function getGlobal(name) {
     this.checkStarted();
     var result = this.evaluate(function _evaluate(name) {
         var result = {};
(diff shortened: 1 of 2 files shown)


istr:
While passing global serializable objects through ```JSON.stringify``` and ```JSON.parse``` is kind of a kludge, this hack to pass non-serializable objects is an even uglier kludge imho.

Closing. Any collaborator: feel free to re-open / or continue discussion.

Issue #520, opened 2013-07-08
New "regex" selector


UlysseM:
I found out there were no way of using real regex for selecting things.
I'm aware of such a feature is not a standard (as opposed to CSS selector and XPATH), but in some case, they can be quite useful.

I wrote a patch that uses a 'regex' selector, which looks pretty much like this:

``` js
var selector = {
  type: 'regex',
  tag: 'a',
  attributes: {
    href: 'yourRegex',
    style: '#33[0-9a-fA-F]{4}'
  }
};
```

If it's not clear enough, the selector will look for the any item of the tag 'a', which contains at least the attributes 'href' and 'style' matching the regex.

istr:
Would you mind adding a test case and turning this into a PR on a separate feature branch that is up-to-date with current master?

BIGjuevos:
I quite like this idea, marking it as part of the 1.2 milestone.  Can @picandocodigo or @UlysseM please pick up the PR so we can include it in 1.2?

kensoh:
@UlysseM and @picandocodigo bumping up this interesting enhancement suggestion.

UlysseM:
I haven't used casperjs in years, but I'll try to bump it this weekend.

kensoh:
Thanks @UlysseM ! Also copying maintainer @jefleponot

kensoh:
Archiving issue using **ARCHIVE** label as CasperJS, like PhantomJS, is not actively maintained for now.

PR #522, opened 2013-07-08
refs #520 - Added 'regex' selector

UlysseM:


kensoh:

Issue #651, opened 2013-10-01
Logging of steps with function names is inconsistent


jpz:
I've found it quite useful to declare non-anonymous step functions, so I can see what step is being executed.  

For a several step definitions, a step function is created which wraps the originally named function, and the "_step" function is logged as being called (which is the wrapping function name declared in the casper.js file.)  

This function name being logged has very little utility to the user.

I've made a few minor changes to casper.js to provide improved logging of function names, on a branch in my fork of the repository.

thanks

kensoh:
Closing this issue for now as it hasn't gotten new replies or discussion for 3+ years. But do reopen if anything!

PR #652, opened 2013-10-01
Improvement to logging of function names. (corrects #651)

jpz:
Let me know if you have any feedback or criticism on the implementation.  

One criticism I had of my approach was whether then() and createStep() could be combined in some way.

Note I didn't get rid of _step in 100% of cases!  For instance - for the forward() and back() functions appear to accidentally have some faults:

1) they are asymmetric (one has a then function parameter, the other doesn't)

2) the then function parameter for the forward() function doesn't seem to be called

So where there were overlapping issues I did not make corrections (yet) - for the purposes of keeping commits separate.

I'm happy for all feedback on approach, variable naming, indentation - please feel free to be direct if you see something you don't like, and I will take it as constructive criticism and guidance on understanding your expectations :)

Cheers

### DIFF
--- a/modules/utils.js
+++ b/modules/utils.js
@@ -792,7 +792,6 @@ exports.unique = unique;
 
 function coalesce(p0, p1, p2, p3, p4, p5, p6, p7, p8, p9)  // jshint maxparams:10
 {
-    /*jshint maxparams:10*/
     "use strict";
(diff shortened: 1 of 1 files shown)


paazmaya:
@jpz any updates on this PR?

PR #686, opened 2013-10-28
casper was unable to listen callbacks for popup pages when window.open is used

jefleponot:
Trouble with popups.
When a "window.open" occurs, casper can't listen to callbacks for popups pages.


jefleponot:
ok n1k0.
But what is a PR ?

jefleponot:
this PR was replaced by  Update popup feature #1118

PR #702, opened 2013-11-12
Update frame code

clochix:
Some update of the frame code:
- methods `childFramesCount`, `childFramesName` and `switchToChildFrame` are deprecated. Replace them with their new synonyms;
- I need to play with the content of an iframe for more than one step, so I exposed the `switchToFrame` method;
- also, I need my clients scripts into the frame, so I injected them;
- I like events, so I added a `frame.changed` event.To be consistent, maybe should I also encapsulate `switchToParentFrame` and `switchToMainFrame`, to also emit this event ?

### DIFF
--- a/bin/casperjs
+++ b/bin/casperjs
@@ -89,7 +89,8 @@ ENGINE_NATIVE_ARGS = []
 ENGINE_EXECUTABLE = ''
 
 CASPER_ARGS = []
-CASPER_PATH = os.path.abspath(os.path.join(os.path.dirname(resolve(__file__)), '..'))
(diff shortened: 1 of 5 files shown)


paazmaya:
@n1k0 next merge candidate?

istr:
@clochix Could you do us a favor and merge back current master to your PR?

jefleponot:
@clochix 

fews years later, we have integrated your code on the PR #1802 .
http://docs.casperjs.org/en/latest/modules/casper.html#switchtoframe  

Thanks a lot.

Issue #912, opened 2014-05-06
XUnit : make failureNode.appendChild to take another property


Fanch-:
What I want to do, is to use the archive artifact option of Jenkins to display the url of a screenshot (test which fails, using casperJs events) in jenkins, in the message of a failure, like that : 
http://hpics.li/3f15d86
That way is quite easy to debug, you click on the url and the image appears in the browser.

Currently, to make a xunit report, both <code>testcase name</code> and <code>failure child -> textNode</code> take the same property : failure.message (if i'm right, I hope so :p ).
So it makes that : http://hpics.li/cd0cbda

What I'd like to have is that : 
http://hpics.li/e906a54

Don't you think it could be useful to add another property in the failure object? Failure.message for testcase name and for example failure.textContent for... failure text content (by default equal to failure.message).

In this way it would be easy to customize our jenkins report : 

	var screenCount = 0;
	if (casper.cli.get("xunit")){
		casper.test.on('fail', function(failure) {
			var capture = '/var/lib/jenkins/jobs/' + myJob + '/workspace/screenshots/fail' + screenCount + '.png';
			casper.capture(capture);
			screenCount++;
			failure.textContent = "Message : " + failure.message + "\nLine : "+ failure.line + "\nCode : " + failure.lineContents 
								+ "\nScreen : " + 'http://-jenkins-/job/' + myJob + '/lastFailedBuild/artifact/screenshots/fail' + screenCount + '.png/';
		});
	}			
Well, we can currently do it with failure.message but then our testcase name is a bit long, it's quite annoying : 
http://hpics.li/585a8ab
The page becomes too large :p  

And I like the fact to customize my testcase name and failure text content independently.

So what do you think?

BIGjuevos:
PR was merged. closing.

PR #920, opened 2014-05-12
refs #912 - add textContent property to the assertion object

Fanch-:
One solution for ref #912.
local selftest Okay.

_Edit_ : Where do I define textContent? ...


kensoh:

Issue #791, opened 2014-01-08
waitFor - Performance - Call testFx immediately before doing setInterval


brandon-bethke-neudesic:
It seems that waitFor will wait for a minimum of 2x the interval. It doesn't call testFx until the setInterval fires and then it waits another interval to check the condition and clear the interval. 

It seems that perhaps waitFor could immediately call the testFx and then if the condition is false, perform the setInterval. This would increase performance.

BIGjuevos:
Related PR is closed.  Closing this issue.

PR #928, opened 2014-05-19
[WiP] refs #791 - improved waitFor performance.

n1k0:
Immediately schedule next step if a `waitFor` condition is already met at the time the inner `waitFor` step is executed.


istr:
The tests are [too old to be restarted](https://travis-ci.org/n1k0/casperjs/jobs/25561499). But I guess we have a good idea why intermittent failures occur from time to time.

It would be great if you could just re-base this PR against current master. This seems to be a very low hanging fruit to resolve an open issue. :-)

kensoh:

PR #1018, opened 2014-08-27
add this.touch function to casper

zezhou:
One of My friends failed to use casperjs to write test case in mobile web page testing, because this web page use touch event to trigger a action, but capserjs only support click event. I wrote following codes to make touch event available in capserjs, and I think a touch event supporting in mobile web page testing is necessary, as the touch event in cellphone is very common. So I commit my patch back to capserjs, and hope it helps others.

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -1928,6 +1928,43 @@ Returns a string representation of current Casper instance::
 
     casper.run();
 
+.. _casper_touch:
+
+.. index:: touch
+
+``touch()``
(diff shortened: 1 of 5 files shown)


ciokan:
Any news on this one being merged?

mickaelandrieu:
I'm not sure this should be added in API, that's why I need @n1k0 review :)

n1k0:
Too much code duplication between the new touch-related methods and the click ones. This patch needs refactoring.

PR #1026, opened 2014-09-04
TestRunner curTestSuiteNum Bug Fix

abelnation:
(this pull request is a cleaned up version of the now-closed PR #937)

Currently curTestSuiteNum is updated for every file that is run, restricting each file to one test suite per file.

This was causing issues in files with multiple sets of tests:
- stepTimeout uses an id of <filenum>_<stepnum>, and stepTimeout stops polling if the id matches
- if a subsequent testSuite runs in the same file, it will have the same testSuiteId, causing potential misfires of stepTimeout

This separates file count from test suite count, eliminating the stepTimeout problem, and allowing multiple testSuites per file.

### DIFF
--- a/modules/tester.js
+++ b/modules/tester.js
@@ -103,6 +103,7 @@ var Tester = function Tester(casper, options) {
     this.currentTestFile = null;
     this.currentTestStartTime = new Date();
     this.currentSuite = undefined;
+    this.currentFileNum = 0;
     this.currentSuiteNum = 0;
     this.lastAssertTime = 0;
     this.loadIncludes = {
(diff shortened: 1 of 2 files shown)


modules/tester.js
```diff
@@ -1136,6 +1137,11 @@ Tester.prototype.done = function done() {
     /*jshint maxstatements:20, maxcomplexity:20*/
     var planned, config = this.currentSuite && this.currentSuite.config || {};
 
+    this.casper.echo(f("done called"), "INFO");
```
istr: I am not sure if we really want to add the INFO echo messages contained in this PR. @n1k0, @hexid, @paazmaya ?

istr:
Looks good, but I would drop the additional echo / INFO messages.

mlalpho:
Can this be merged (minus debug output) ?

jefleponot:
ok
I will rebase the 2 PR. 
@mickaelandrieu , do you agree ?

kensoh:

PR #1071, opened 2014-10-29
Ensure headers["Content-Type"] exists before comparing string.

dtarda:


paazmaya:
Might be, with a broken server or with a mockup data...

corford:
We were getting "TypeError: undefined is not an object (evaluating 'settings.headers["Content-Type"].match')" when POSTing to a certain site (using the master branch of Casper and phantomjs-2.0.1).

I know phantomjs2 isn't officially supported yet but FYI this patch fixed the issue.

mickaelandrieu:
Thanks ;)

PR #1110, opened 2014-12-16
Make concise work

Ajnasz:
--concise parameter were not passed to tester and there was a check if it's true, which should check check to false I guess.

### DIFF
--- a/modules/tester.js
+++ b/modules/tester.js
@@ -1063,7 +1063,7 @@ Tester.prototype.begin = function begin() {
         config = getConfig([].slice.call(arguments)),
         next = function() {
             config.test(this, this.casper);
-            if (this.options.concise)
+            if (!this.options.concise) {
                 this.casper.echo([
                     this.colorize('PASS', 'INFO'),
                     this.formatMessage(description),
@@ -1071,6 +1071,7 @@ Tester.prototype.begin = function begin() {
                                     config.planned,
                                     config.planned > 1 ? 's' : ''), 'INFO')
                 ].join(' '));
+            }
         }.bind(this);
 
     if (!this.options.concise)
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -179,7 +179,9 @@ var Casper = function Casper(options) {
             throw new CasperError('casper.test property is only available using the `casperjs test` command');
         }
         if (!utils.isObject(this._test)) {
-            this._test = tester.create(this);
+            this._test = tester.create(this, {
+                concise: this.cli.get('concise')
+            });
         }
         return this._test;
     });


paazmaya:
Not exactly sure how `concise` is supposed to affect the tests, but since by default it is `false` the given changes looks legitimate.

mickaelandrieu:
@Ajnasz would you mind to rebase master branch please ? 
Sounds good, but unmergeable.

Ajnasz:
I rebased and force pushed it, but one build failed. I'm not sure, the issue is related.

mickaelandrieu:
I relaunch the build, ping /c @paazmaya can you review this one too ?

I'll merge with your :+1:

modules/tester.js
```diff
@@ -1063,7 +1063,7 @@ Tester.prototype.begin = function begin() {
         config = getConfig([].slice.call(arguments)),
         next = function() {
             config.test(this, this.casper);
-            if (this.options.concise)
+            if (!this.options.concise)
```
paazmaya: Perhaps curly brackets could be used here, as it seems to be the case with the rest of the file

paazmaya:
Looks good, so all I can say is :+1:

modules/tester.js
mickaelandrieu: For consistency, I agree.

Then, let's go for merge (I will probably update docs)

paazmaya: I think the merge could be done now, and the curly brackets consistency added later.

Ajnasz: Well, I can add braces, but this file has some other if statements without braces.
What is your current styleguide? Braces for every if statement?

paazmaya: Braces for everything would seem to be the future approach.

Ajnasz: I added the braces, but now jshint breaks the build :)
Should I go through on those errors?

paazmaya: It is about the given file having too many statements.

I am currently working on #1133 which touches this, so you can leave it as it is.

mickaelandrieu:
```
OK
jshint .
modules/tester.js: line 1031, col 40, This function has too many statements. (18)
```
this is related.

Ajnasz:
I reverted most of the braces I added, because the jshint version looks buggy for me, see the commit message in 27ac95e346edb

mickaelandrieu:
Ok. Squash your commits and I'll merge it: we shouldn't waste "too much" time to please js(h/l)int...

Ajnasz:
Squashed it.

mickaelandrieu:
Merged, thank you @Ajnasz !

PR #1113, opened 2014-12-24
Added warning about overriding 'page' property

robi50:
Sorry for my bad english. I will be glad if someone could correct my changes. (funny but including this apology note)


paazmaya:
Looks legitimate

PR #1124, opened 2015-01-12
Change type for assertNotVisible

keathley:
Seeing the 'assertVisible' type in the test report can be very confusing.  I've changed the type to 'assertNotVisible' to match the actual assertion name.


n1k0:
Thanks.

Issue #1085, opened 2014-11-06
casper.visible doesn't report true for "display: flex"


ryanfields:
I'm not sure if this is an issue with casper's code, or with PhantomJS itself.  It doesn't appear that the casper.visible test works for elements with a display setting of ```flex```.

paazmaya:
I started doing PR for this, but added tests seem to pass without any changes to `clientutils.js`....

Do I need to do something else besides running `make selftest` for testing that?

(and of course creating the fixture with `display: flex` and `inline-flex`)

nchase:
looks like this was fixed by #1668, can this be closed?

PR #1136, opened 2015-01-28
Handle flex and inline-flex for reading visible state

paazmaya:
Regarding #1085, this adds the checks and tests for getting visible state for elements that have `display: inline-flex` or `display: flex`.


mickaelandrieu:
I'm not sure that the webkit version used by phantomjs handle this kind of css property.

This need more investigation on webkit4 (phantomjs 1.x) and webkit 5.

istr:
@paazmaya Please rebase against current master so we can see if your test passes for the supported engines in master.

paazmaya:
Have I added enough commits on this.... ?

:neckbeard: will be creating a new PR

PR #1138, opened 2015-01-28
Added link to Gitter

rdpanek:
Please @n1k0 create room at Gitter :+1:


mickaelandrieu:
I'm not sure ... I agree to get a gitter *but* I think this kind of decision is more up to @n1k0.

paazmaya:
Perhaps this can be re-evaluated when the project moved under `casperjs` organisation?

slang800:
:-1: - Gitter has no place in FOSS development. The last thing we need is another proprietary chat network, especially since we already have ubiquitous IRC support.

n1k0:
IRC is good enough.

PR #1139, opened 2015-01-28
Workaround on phantomjs to support phantomjs 1.9.8

mickaelandrieu:
fixed Unsafe Javascript attempt, see https://github.com/ariya/phantomjs/issues/12697 for workaround suggested.

### DIFF
--- a/bin/bootstrap.js
+++ b/bin/bootstrap.js
@@ -87,18 +87,22 @@ CasperError.prototype = Object.getPrototypeOf(new Error());
         return;
     }
 
+    function __exit(statusCode){
+        setTimeout(function() { phantom.exit(statusCode); }, 0);
+    }
+
     function __die(message) {
         if (message) {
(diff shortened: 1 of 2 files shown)


paazmaya:
@mickaelandrieu should there be a comment in the new code stating that it is needed only for the given workaround?

Also, I don't quite understand why there was a PR which was very soon merged by the same person. Please correct me if I am wrong, but wouldn't the point of having a PR is for possible comments and someone else to merge it?

mickaelandrieu:
Hi,
you are right in case there is enough active contributors to take review and merge.

As you may now, CasperJS build is broken since months and as stated in link in my PR, the workaround is known and already used in some others tools from phantomjs family.

Afaik only @n1k0 and I can merge and he is busy since and for months.

I will make a PR for future contribution, I'm glad to see new active contributors :+1:

PR #1154, opened 2015-02-12
fixed phantom.args for phantomjs 2.x

goggelz:
@n1k0 this works for me for the phantomjs 2.0 version.
I'm not sure, if it would work for the 1.x version, so that's why I checked it that early only to set the phantom.args

### DIFF
--- a/bin/bootstrap.js
+++ b/bin/bootstrap.js
@@ -40,7 +40,16 @@ if ('process' in this && process.title === "node") {
 // phantom check
 if (!('phantom' in this)) {
     console.error('CasperJS needs to be executed in a PhantomJS environment http://phantomjs.org/');
+} else {
+    if (phantom.version.major === 2) {
+        //setting other phantom.args if using phantomjs 2.x
+        var system = require('system');
+        var argsdeprecated = system.args;
+        argsdeprecated.shift();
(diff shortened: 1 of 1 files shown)


n1k0:
Ah, my bad I was unclear; you need to add a new entry for PhantomJS 2 to the Travis build matrix (basically after [this line](https://github.com/n1k0/casperjs/blob/master/.travis.yml#L51)), so we can ensure that everything is fine.

n1k0:
> there are no binary files for linux packed for PhantomJS 2

Wat. Damn, that sucks. So forget about it, I guess. Grmbl, snap.

In other news, the build is green, which makes me confident we can safely merge that. We can always revert later if people complain ;)

goggelz:
Great :)

msankhala:
Please update the same for beta so one can install with brew.

kaiye:
I just install from brew, met the problem, and solved by overwrite the file : /usr/local/Cellar/casperjs/1.1-beta3/libexec/bin/bootstrap.js, thanks a lot :+1:

alexserver:
Can you please release a new beta or update the latest one ? so we can move forward with the error ?
@goggelz thanks for the fix !

Issue #1165, opened 2015-02-24
Python bin doesn't support path with spaces


Philoozushi:
When python bin file is in a path containing spaces
(typically with npm, e.g. `~/My Project/node_modules/casperjs/bin/casperjs`),
the `CASPER_COMMAND` becomes `['~/My', 'Project/node_modules/casperjs/bin/casperjs']`
because of `ENGINE_EXECUTABLE.split(' ')` in line 127.
Obviously, we get a `not found` error.

Philoozushi:
Yeah, several tries for the fix :stuck_out_tongue:
This is my first time trying to fix / do pull requests with github, and trying to respect the guidelines.

paazmaya:
Fixed via merge 922262747c7ec972cf48a4df5ce7942774dbbcbd

Issue #269, opened 2012-11-06
Windows batch script unexpectedly splits arguments and does not support path with spaces


langholz:
Summary:
-------------------------
The windows batch script (batchbin\casperjs.bat) breaks the execution of scripts when there are spaces provided in the PATH environment variable or command line arguments are passed with values.

Details:
-------------------------
*Version:*
PhantomJS - Version 1.7
CasperJS - Version 1.0.0-RC4
Windows 8

<dl>
  <dt>Paths with spaces are not supported:</dt>
  <dd>Repro steps:</dd>
  <dd><pre>
> set path=%path%;C:\tmp\Test Folder\phantomjs
> set path=%path%;C:\tmp\Test Folder\casperjs\batchbin
> casperjs dumpargs.js</pre></dd>
  <dd>Output:</dd>
  <dd><pre>Can't open 'C:\tmp\Test'</pre></dd>
  <dt>Invalid splitting of arguments with values:</dt>
  <dd>Repro steps:</dd>
  <dd><pre>
> set path=%path%;C:\tmp\Test Folder\phantomjs
> set path=%path%;C:\tmp\Test Folder\casperjs\batchbin
> casperjs dumpargs.js --test="wo0t" --ph34r</pre></dd>
  <dd>Actual output:</dd>
  <dd><pre>
Casper CLI passed args:
[
    "wo0t"
]
Casper CLI passed options:
{
    "casper-path": "C:\\tmp\\scripts\\casperjs\\batchbin\\..\\",
    "cli": true,
    "test": true,
    "ph34r": true
}</pre></dd>
  <dd>Expected output:</dd>
  <dd><pre>
Casper CLI passed args:
[]
Casper CLI passed options:
{
    "casper-path": "C:\\tmp\\scripts\\casperjs\\batchbin\\..\\",
    "cli": true,
    "test": "wo0t",
    "ph34r": true
}</pre></dd>
</dl>

*dumpargs.js:*
<pre>
var casper = require("casper").create();
casper.echo("Casper CLI passed args:");
require("utils").dump(casper.cli.args);
casper.echo("Casper CLI passed options:");
require("utils").dump(casper.cli.options);
casper.exit();
</pre>

Philoozushi:
For you Windows guys (I'm not one) :
You should try removing `Split(' ')` in `src/casperjs.cs`, line 45, then compile and try Casper again.

PR #1166, opened 2015-02-24
refs #1165 - fixed python bin for paths w/ spaces

Philoozushi:
It's now better and works with `[ENGINE_EXECUTABLE]`.

### DIFF
--- a/bin/casperjs
+++ b/bin/casperjs
@@ -124,7 +124,7 @@ for arg in SYS_ARGS:
     if not found and arg_name != 'engine':
         CASPER_ARGS.append(arg)
 
-CASPER_COMMAND = ENGINE_EXECUTABLE.split(' ')
+CASPER_COMMAND = [ENGINE_EXECUTABLE]
 CASPER_COMMAND.extend(ENGINE_ARGS)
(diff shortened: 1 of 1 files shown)


bin/casperjs
```diff
@@ -124,7 +124,7 @@ for arg in SYS_ARGS:
     if not found and arg_name != 'engine':
         CASPER_ARGS.append(arg)
 
-CASPER_COMMAND = ENGINE_EXECUTABLE.split(' ')
+CASPER_COMMAND = [ENGINE_EXECUTABLE]
```
n1k0: Damn, seems so obvious… how could we missed that -_-

Philoozushi: I'm not a Windows guy, but it should be the same for the win exe (src/casperjs.cs, line 145), and maybe solve issue #269?

n1k0: Sounds like so. Unfortunately I don't have a clue how to build this thing.

n1k0: Ah, there's a `compile-dotNET` command in the Makefile, fortunately.

n1k0:
Thanks for contributing.

PR #1183, opened 2015-03-14
Fixes #1182 - Allow waitFor timeout exception to propagate

### LINKED ISSUE #1182: Exceptions thrown by waitFor timeout callback do not get propagated
Any exception thrown by the onWaitTimeout callback in waitFor will not be propagated. This happens because there is a finally block which contains a return statement which prevents the error from being thrown.

jsherin:
Removed the finally block and moved the code that is not intended to execute into an else block

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -2121,14 +2121,13 @@ Casper.prototype.waitFor = function waitFor(testFx, then, onTimeout, timeout, de
                     if (!self.options.silentErrors) {
                         throw error;
                     }
-                } finally {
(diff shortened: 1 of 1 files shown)


n1k0:
Wow, the finally/return thing was so odd, thanks for fixing.

jsherin:
ya, I thought the behaviour was a javascript bug when a found it but the last example kind of mentions it in: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

Issue #1210, opened 2015-04-14
Doc for thenOpenAndEvaluate does not match code


francois2metz:
In the doc, the signature for `thenOpenAndEvaluate` is `thenOpenAndEvaluate(String location[, Function then[, arg1[, arg2[, …]]])`

But in the code, extra args are not used. This behavior should be similar to `thenEvaluate`.

PR #1211, opened 2015-04-14
Pass extra args to thenEvaluate in thenOpenAndEvaluate function

francois2metz:
See #1210


n1k0:
LGTM, thanks for contributing.

PR #1215, opened 2015-04-16
fix xpath error message

scottlabs:
Hi,

When doing something like this:

            var selector = {
                type: 'xpath',
                path: '//*[@class="plop"]'
            };
            var info = casper.getElementInfo(selector);

Where selector is an element that does not exist, the error reported looks like this:

`CasperError: Cannot get informations from [object Object]: element not found.`

It would be very helpful if the error reported the XPath selector path in the error, instead of `[object Object]`.


n1k0:
Correct [usage](http://docs.casperjs.org/en/latest/selectors.html#index-2):

```js
var x = require('casper').selectXPath;
…
var info = casper.getElementInfo(x('//*[@class="plop"]'));
```

PR #1217, opened 2015-04-17
Fix fetchText() returning "undefined" for empty elements

### LINKED ISSUE #1206: casper.fetchText() will often return the string "undefined"
A PR was merged on [Nov 3rd, 2014] (https://github.com/n1k0/casperjs/commit/bb706acdf38f30d2380ebd3897ba1d4cee74430c) from @andy-g that updated the fetchText() function to [also check the element's value attribute](https://github.com/n1k0/casperjs/commit/bb706acdf38f30d2380ebd3897ba1d4cee74430c) (when the inner text/html is empty) -- which was a nice improvement! However, if the OR logic makes it to `element.value` and the value attribute is undefined (as it often is for non-input elements), then the result ends up being the string "undefined", instead of an empty string.

For example, if you have `<p id="test"></p>`, the result of `casper.fetchText('#test')` will be the string "undefined", not "". (This appears to be the result of JavaScript converting an undefined value to a string.)

A quick fix would be to amend the logic like so:

`text += element.textContent || element.innerText || element.value || '';`

I'm happy to create a PR to fix this and/or a test proving the above, I just wanted to run it by you guys first.

justinklemm:
This resolves #1206 

A PR was merged on [Nov 3rd, 2014] (https://github.com/n1k0/casperjs/commit/bb706acdf38f30d2380ebd3897ba1d4cee74430c) that updated the fetchText() function to [also check the element's value attribute](https://github.com/n1k0/casperjs/commit/bb706acdf38f30d2380ebd3897ba1d4cee74430c) (when the inner text/html is empty) -- which was a nice improvement! However, if the OR logic makes it to `element.value` and the value attribute is undefined (as it often is for non-input elements), then the result ends up being the string "undefined", instead of an empty string.

For example, if you have `<p id="test"></p>`, the result of `casper.fetchText('#test')` will be the string "undefined", not "". (This appears to be the result of JavaScript converting an undefined value to a string.)

This PR fixes the issue and includes a test for it.

### DIFF
--- a/modules/clientutils.js
+++ b/modules/clientutils.js
@@ -230,7 +230,7 @@
             var text = '', elements = this.findAll(selector);
             if (elements && elements.length) {
                 Array.prototype.forEach.call(elements, function _forEach(element) {
-                    text += element.textContent || element.innerText || element.value;
+                    text += element.textContent || element.innerText || element.value || '';
                 });
             }
             return text;
--- a/tests/suites/casper/fetchtext.js
+++ b/tests/suites/casper/fetchtext.js
@@ -27,3 +27,14 @@ casper.test.begin('fetchText() handles HTML entities', 1, function(test) {
         test.done();
     });
 });
+
+casper.test.begin('fetchText() handles empty elements', 1, function(test) {
+    casper.start().then(function() {
+        this.setContent('<html><body></body></html>');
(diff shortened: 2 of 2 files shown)


n1k0:
LGTM, thanks for contributing.

Issue #1061, opened 2014-10-16
Cannot dispatch Mouseup event on nonexistent selector (it did exist though)


cablegunmaster:
Basically it never recovers from going down by a disappearing Selector.
Click removes the selector. so it can't bounce up again.

Code:
test.assertExist(".selectize-dropdown-content .option:nth-child(2)"); //true.
this.click('.selectize-dropdown-content .option:nth-child(2)'); //gives error on mouseup event.

Stack trace:
"Cannot dispatch mouseup event on nonexistent selector"

hotfix: try catch surrounding that event and force a this.mouse.up('body');

Gyomu:
I add the same issue on "click" event though.
In my case I'm testing a GWT application and firing up mousedown then mouseup events are enough.

When we call this line : `success = success && this.mouseEvent('click', selector);`
a test is done in the mouseEvent method that triggers an exception :
```javascript
if (!this.exists(selector)) {
        throw new CasperError(f("Cannot dispatch %s event on nonexistent selector: %s", type, selector));
    }
```
In my specific case the selector is long gone when hitting this line, thus ending up with an error being returned.
The following modification solved the problem for me (which might not be the best way to solve this issue)
```javascript
if (this.exists(selector)) {
    success = success && this.mouseEvent('click', selector);
  }
```

Llorx:
This bug is simply because the item has a listener to the event mousedown removing itself from the DOM, so when it tries to call mouseup, the element does not exists.

kensoh:
Seconding @Llorx, @cablegunmaster sounds like have to use mouse.down method instead. The default click will activate mouse.down, mouse.click and mouse.up I think..

kensoh:
Seconding @Llorx, @cablegunmaster sounds like have to use mouse.down method instead for this edge case. The default click will activate mouse.down, mouse.up, and mouse.click I think..

kensoh:
Thanks @cablegunmaster for raising this. Closing this issue as mouse.down method can be used instead of the default down+click+up when invoking this.click. In any case do ping on GitHub here if you still have the issue. Or email at ken@tebel.org

PR #1220, opened 2015-04-24
Fix the cannot dispatch click event or mouseup event on nonexistent selector issue

Gyomu:
see #1061


istr:
A test for this would be great.

kensoh:

PR #1247, opened 2015-06-10
Added getFieldValue to Casper prototype

a-West:
Made it more convenient to use the callUtils getFieldValue by adding it to the Casper prototype. Allows for cleaner scripting.


kensoh:

Issue #1252, opened 2015-06-17
casper differences with phantomjs and slimerjs


jefleponot:
My casperjs program is fully functional with phantomjs but it must be also compatible with slimerjs.
Many trouble occur when i used it with Slimerjs. I begin to list these troubles.

0 / General exit : 
------------------

- phantomjs exit method is synchronous
- slimerjs  exit method is asynchronous

so if there are another script lines after the exit code, slimerjs executes them like 'display, log ... echo'

1/ phantom.exit :
------------------

<pre>
var casper = require('casper').create();
...
if (!runningConditions) phantom.exit();
...
</pre>

Results : 
- phantomjs ended normally
- slimerjs ended with an empty red bar (because of an opening error event)

2 / casper.exit :
------------------

<pre>
var casper = require('casper').create();
...
if (!runningConditions)  casper.exit(0);
...
casper.start();
</pre>

Results : 
- phantomjs ended normally
- slimerjs ended with core dumped (segmentation error) because of casper.start() command.

to be continue :-)

PR #1253, opened 2015-06-17
do not display empty error message

jefleponot:
#1252


mickaelandrieu:
@jefleponot as this part of casperjs is actualy disabled, I don't see the value of this contribution (for now).

What do you think ? Could be vey interesting if we can re enable this part, and then enjoy your enhancement.

Mickaël

PR #1261, opened 2015-07-02
Docs: Update optional params to evaluateOrDie

Dudemullet:
evaluateOrDie is already capable of passing a custom status code to die. Just updating the docs to reflect this.


mickaelandrieu:
Hi,

you are right.

Thank you @Dudemullet 

PS: tests fails cause we have moved casperjs repositoru to his own organization.

PR #1268, opened 2015-07-12
Slight rewording for clarification

dxuehu:


rippo:
I think this slightly makes more sense

"nor does it require native NodeJS modules"

dxuehu:
updated

istr:
This was addressed and clarified in 679c93a3452117115f1e2e136b194199057b5e4f. Closing.

Issue #1275, opened 2015-07-19
CasperError: Errors encountered while filling form: RangeError: Array size is not a small enough positive integer.


Lucas-C:
Hello,

This is a bug report : I identified the root cause of the error mentioned in this issue title.
When trying to fill a form, I was passing an integer as a password and the following code crashed because `value.length` evaluated to `undefined`:
https://github.com/n1k0/casperjs/blob/master/modules/clientutils.js#L828

I guess either `value` should be checked or converted to string at some point beforehand,
or this code line should handle non-string values.

    $ casperjs --version && phantomjs --version
    1.1.0-beta3
    1.9.8

timscott:
I just got this. Workaround?

Lucas-C:
Convert the integer to string:

    ''+42

istr:
Merged. Thanks for your PR.

PR #1289, opened 2015-08-11
Bug fix for #1275 : adding missing conversion to string

Lucas-C:


istr:
Would you mind adding a simple test to ```tests/suites/casper/formfill.js``` that references the original issue #1275?

I will merge this when there's a test included demonstrating that this issue is resolved by this PR (even though the fix itself is trivial).

Thanks.

(This points to a general problem with "type safety" in ```setField``` and and thus ```fill```. More tests and more robust "convert to string" policy would be nice to have.)

Lucas-C:
Done !

istr:
@paazmaya,  @hexid Please have a short look. I would like to merge this.

mickaelandrieu:
:+1: this make sense

istr:
@mickaelandrieu Ok, thank you. Sorry that I missed to address you personally, too.

mickaelandrieu:
@istr it's OK,  I'm not very active ;)

Issue #1290, opened 2015-08-11
Infinitely loops when calling casper.warn in on(fail) handler


hivivo:
The following code is a minimum test case:

``` javascript
casper.test.on('fail', function onFail(failure) {
  casper.warn('this will cause infinite loops');
});

casper.test.begin('Demo suite', function(test) {
    casper.open('Does not exist');
});
```

When trying to run this test, the result would be endless like:

```
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
FAIL TypeError: Attempted to assign to readonly property.
...
```

I have figured out the root cause and will push a fix soon.

kensoh:
Hi @hivivo I've just checked and it looks like your changes are now in tester.js :)

kensoh:
Thanks @hivivo! Closing this issue for now, but not sure why your commit isn't shown as merged. Though your changes are now in tester.js code.

PR #1291, opened 2015-08-11
refs #1290: Infinitely loops when call casper.warn()

lovivo:
All the usage is to treat warning as a string. Trying to assign a property to a string would trigger an uncaughtError exception. Besides, xunit.js also uses warnings as a string array to join them together.

### DIFF
--- a/modules/tester.js
+++ b/modules/tester.js
@@ -1958,11 +1958,10 @@ TestCaseResult.prototype.addSkip = function addSkip(skipped, time) {
 /**
  * Adds a warning record.
  *
- * @param Object  warning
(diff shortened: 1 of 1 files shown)


istr:
[My 2.0.0 PR](https://github.com/n1k0/casperjs/pull/1352) addresses / fixes this issue, too.
See [this commit](https://github.com/istr/casperjs/commit/146e93e0b5d9f8b706357495721fe50f57761170).

istr:
Closed with [PR 1352](https://github.com/n1k0/casperjs/pull/1352).

PR #1306, opened 2015-08-27
thenEvaluate resolves promise with result of evaluate.

alexbiehl:


alexbiehl:
Closing this, didn't realize, that then is ment in a promise composing way.

PR #776, opened 2013-12-19
Print out Casper help on each travis build

hexid:
This will ensure that the correct engine version is being used.

ref #775

PR #1310, opened 2015-09-02
PhantomJS v2.x support (maintaining backwards-compatibility)

hexid:
This PR should allow PhantomJS v2.x to pass most of the selftests.
There are still some failures that appear to be caused by [regressions](https://github.com/ariya/phantomjs/issues/12506) that should be fixed with [PhantomJS v2.0.1](https://github.com/ariya/phantomjs/issues/12902).

This also fixes a few issues that I found while working around failing tests.

A few points still need to be addressed before this can be considered ready:

 - [x] Travis needs to be setup for PhantomJS v2.0.0
 - [x] SlimerJS stalling on CoffeeScript tests (only in v0.9.6 - fixed in master)
 - [x] Laucher needs to be fixed so SlimerJS actually gets tested on Travis
 - [ ] Fix tests that should have been failing for SlimerJS
 - [x] Fix tests that should have been failing for PhantomJS v1.8.2 (support dropped)

If anyone has any comments or concerns, please let me know.

/cc @n1k0 @mickaelandrieu


tests/suites/require.js
```diff
@@ -15,11 +15,15 @@ casper.test.begin('Javascript module loading', 1, function(test) {
 
 casper.test.begin('CoffeeScript module loading', 1, function(test) {
     var csmod;
-    try {
-        csmod = require(fs.pathJoin(modroot, 'csmodule'));
-        test.assertTrue(csmod.ok, 'require() patched version can load a coffeescript module');
-    } catch (e) {
-        test.fail('require() patched version can load a coffeescript module');
+    if (Object.keys(require.extensions).indexOf('.coffee') !== -1) {
```
n1k0: Nit: `".coffee" in require.extensions`

modules/utils.js
```diff
@@ -464,7 +464,11 @@ exports.isHTTPResource = isHTTPResource;
 function isJsFile(file) {
     "use strict";
     var ext = fileExt(file);
-    return isString(ext, "string") && ['js', 'coffee'].indexOf(ext) !== -1;
+    var valid = Object.keys(require.extensions).map(function (val) {
+        return val.replace(/^\./, '');
+    });
+    valid.splice(valid.indexOf('json'), 1);
```
n1k0: Nit:
```js
var valid = Object.keys(require.extensions).map(function(ext) {
    return ext.replace(/^\./, '');
}).filter(function(ext) {
    return ext === "json";
});
```

tests/suites/require.js
n1k0: Also, as there's a single specific test here it's probably worth moving that if statement around the `casper.test` call itself, as it's been done elsewhere already. Adding a comment to explain why we're excluding coffee files might be worthwhile too.

n1k0:
> SlimerJS appears to be stalling on the CoffeeScript tests (preexisting?)

Which version are you using?

> Laucher needs to be fixed so SlimerJS actually gets tested on Travis

Yeah, it's been bringing false positives for too long. While you're at it, could you handle this as well?

Thanks for your help.

hexid:
> Which version are you using?

    $ slimerjs --version
    Innophi SlimerJS 0.9.6, Copyright 2012-2015 Laurent Jouanneau & Innophi

I can check into why this is happening, but i may not get to it until this weekend.

---

Right now Travis is using the `ENGINE_EXECUTABLE` environment variable, which isn't being used by either the python or the dotNet launchers.
Some options that I'm considering are to:
 * Update both launchers to use this env variable in addition to the `PHANTOMJS_EXECUTABLE` and `SLIMERJS_EXECUTABLE` (I'm assuming that these variables should override the generic `ENGINE_EXECUTABLE` if they are present)
 * Update the Travis config to set the correct env variable

The second option may actually end up being more work in the long run, but I'll investigate both.

---

With either of those options, we will need to either update both launchers to use a `CASPERJS_ENGINE` env variable as an alternative to the `--engine=` argument or allow the makefile to take arguments so that the engine can be specified.

I'm leaning towards the first option because it would make the feature available outside of our testing.
It shouldn't be much more than changing [line 86 of the python launcher](https://github.com/n1k0/casperjs/blob/master/bin/casperjs#L86) to:

```python
ENGINE = os.environ.get('CASPERJS_ENGINE', 'phantomjs');
```

n1k0:
> I'm leaning towards the first option because it would make the feature available outside of our testing.

Makes sense, go for it.

hexid:
I forgot that the `ENGINE_FLAGS` variable is appended to `ENGINE_EXECUTABLE` 

    ENGINE_FLAGS="--local-to-remote-url-access=yes --ignore-ssl-errors=yes"'
    ...
    export ENGINE_EXECUTABLE="$ENGINE_EXECUTABLE $ENGINE_FLAGS"

I'll look into this and hopefully get a patch out tonight.

---

Also, this patch should replace #776

hexid:
My last commit was just a hunch and it looks like I was right; the `ENGINE_EXECUTABLE` variable is out of scope of the `script` section in the Travis tests.
I'll try exporting the variables, so the lines would look like:

    if [[ $ENGINE == 'phantomjs' ]]; then
      export ENGINE_EXECUTABLE="engine/bin/phantomjs";
    elif [[ $ENGINE == 'slimerjs' ]]; then
      export ENGINE_EXECUTABLE="engine/slimerjs";

@n1k0 Would you like me to revert the changes that I made in b042eba?

n1k0:
> Would you like me to revert the changes that I made in b042eba?

No that's useful, so we know the executable env configuration.

hexid:
@n1k0 @mickaelandrieu 

I've been looking into the failing tests and it appears as though the PhantomJS v1.8.2 tests are failing due to the page.onResourceError event not having been implemented until v1.9.0 (https://github.com/ariya/phantomjs/issues/10954). Support was also not added to SlimerJS until v0.9.0 (https://github.com/laurentj/slimerjs/issues/92).

With these versions being more than 2 years old, it may be time to remove support for them. Thoughts?

    Test file: /home/travis/build/n1k0/casperjs/tests/suites/casper/resources.js    
    # Basic resources tests
    PASS Basic resources tests (5 tests)
    PASS only one resource found
    PASS two resources found
    PASS phantom image found via test RegExp
    PASS phantom image found via test Function
    PASS phantom image found via test String
    # "resource.error" event
    PASS "resource.error" event (3 tests)
    Test server url not found: /tests/site/non-existant.html (file: /home/travis/build/n1k0/casperjs//tests/site/non-existant.html) warning
    ⚠  "resource.error" event: 3 tests planned, 0 tests executed

n1k0:
> With these versions being more than 2 years old, it may be time to remove support for them. Thoughts?

Definitely.

brunowego:
:+1:

istr:
Great! Thanks for the fundamental work to prepare phantomjs2 support. :+1:

PR #1315, opened 2015-09-13
Update installation.rst

JZL:
If I just run `npm install -g casperjs` I get an error about no compatible targets (full message at http://pastebin.com/zAdpTtdB)

The way to fix this is add the @VERSION to the npm but some people might be too scared of the error message to really understand it. I think the clearer way is to show the possible examples from the doc.

I'd like to include a little information about which version (beta2 or beta3) should be chosen but I don't know enough about the differences.

Cheers!

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -49,10 +49,17 @@ Installing from npm
 
 .. versionadded:: 1.1-beta3
 
-You can install CasperJS using `npm <http://npmjs.org/>`_::
+You can install CasperJS using `npm <http://npmjs.org/>`_:
 
-    For beta3: $ npm install -g casperjs@1.1.0-beta3
-    For beta2: $ npm install -g casperjs@1.1.0-beta2
+- For most users:
+
+    $ npm install -g casperjs
+    
+- If want specific version:
+
+    - For beta3: $ npm install -g casperjs@1.1.0-beta3
+
+    - For beta2: $ npm install -g casperjs@1.1.0-beta2
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


paazmaya:
Perhaps the documentation could contain a link to the relevant section of npm-install and have first command example without the version, while the second would be `beta-3`.

JZL003:
I agree that would work; really I thing anything will. I just feel like the current documentation is confusing (I know it took me a few minutes to fix it) and the first install is the point where new users are most likely to quit

docs/installation.rst
```diff
@@ -51,7 +51,8 @@ Installing from npm
 
 You can install CasperJS using `npm <http://npmjs.org/>`_::
 
-    $ npm install -g casperjs
```
istr: Please re-enable this line in your PR, because we want to address this issue with casperjs@1.1.0-beta4.

mickaelandrieu: @istr is there something to do to make this command choose 1.1.0-beta4 ? I dont know how node package manager work.

JZL: @istr I do not understand. If I understand your PR correctly for 1.1.0-beta4 (#1365) `npm install -g casperjs` should work as expected/be addressed. So there is no need for my PR?

istr:
@JZL I think there still is the need to point out how to install a specific, older (not the latest) version. So this PR is still valid and relevant.

JZL:
@istr OK, does the newest commit solve this then?

istr:
Merging. Further updates to this document will follow in a dedicated 1.1.0-beta4 branch.

PR #1320, opened 2015-09-20
added phantomjs2 to travis

mickaelandrieu:
Now the jshint/lint/... issues are fixed, this is an update of travis configuration file to launch testsuite against phantomjs 2.x branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
     fi
 
 before_script:
-  - "npm install -g eslint@v1.2.1"
+  - "npm install -g eslint"
 
 script:
   - "make $MAKE_TEST_COMMAND"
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
ping @n1k0 what do you think ? the build is full green, this cant be possible -_-

I'm not sure we should merge this one because people will think that casperjs 1.1-x is compatible with phantomjs 2 and we know this is not true...

hexid:
@mickaelandrieu Can you take a look at the last build for #1310?
That build should be testing against the correct engine versions, however there is still some work to be done to get all of those versions passing.

I'm working on getting SlimerJS v1.8.4 and v1.8.5 working at the moment, and hope to have them ready by the end of the week.

mickaelandrieu:
:+1:

PR #1346, opened 2015-11-16
Allow form fill to find select element values by their option elements text as well as value

andy-g:
Currently, using the form fill helper requires one to specify the values for a select element, but at times we may not know the value, but we will know the text of the option we want to select.

This will also make the test more legible, as the value specified in the test will be the same as the value the user would look for and choose in the browser.

I also added url matching to test that the correct multi-select values are being submitted

### DIFF
--- a/modules/clientutils.js
+++ b/modules/clientutils.js
@@ -886,7 +886,7 @@
                         });
 
                         // If the values can't be found, try search options text
-                        if (field.value == "") {
+                        if (field.value === "") {
                           [].forEach.call(field.options, function(option) {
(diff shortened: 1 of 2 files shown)


andy-g:
From what I can see that last Travis result is an error (`Couldn't resolve host 'github.com'`) as opposed to the previous eslint failure. Would someone be able to kick off this run again? Thanks

paazmaya:
Kicked :walking:

andy-g:
Thanks

n1k0:
Great patch, thanks.

Issue #1348, opened 2015-11-19
It seems that travis test suite never actually calls slimerjs for selftest


istr:
Even though ```ENGINE_EXECUTABLE``` is exported correctly within ```.travis.yml``` setting this to slimerjs actually never hits the call ```bin/casperjs selftest```.

This is because
1. bin/casperjs sets the default engine to "phantomjs" (bin/casperjs line 86)
2. the Makefile does not pass the engine to bin/casperjs via --engine= option (this currently is the ONLY way to change the engine for bin/casperjs, see bin/casperjs line 86 ff)

I guess this is the root cause why [current master builds pass](https://travis-ci.org/n1k0/casperjs/jobs/91374814) and [this build fails](https://travis-ci.org/n1k0/casperjs/jobs/81889164) as well [as this build](https://travis-ci.org/n1k0/casperjs/jobs/81889164).

paazmaya:
Would you like to create a PR that ensures the use of `slimerjs` in the specific cases?
I would gladly merge it, even if those tests are failing.....

istr:
Yes, I am [on the way to actually fix](https://github.com/istr/casperjs/commit/6b1589d94b8946cde0efefecd8e8b5765ee41e74) the failing tests...
[Tests up to suites/casper/reload.js already pass for slimerjs](https://travis-ci.org/istr/casperjs/jobs/92129988).

If all goes well then a PR for 2.0.0, 2.0.1 pre-relase support and passing slimerjs tests will be in place tomorrow.

istr:
Added PR https://github.com/n1k0/casperjs/pull/1349

paazmaya:
Thank you, merged.

PR #1349, opened 2015-11-19
Actually use the correct engine from env in Makefile.

istr:
This addresses #1348.


paazmaya:
Thank you sir!
Strange that it has been like this for some time...

I am merging this for the sake of having visibility for SlimerJS support.

PR #1359, opened 2015-11-25
Clarify what to do if in need of help.

istr:
Try to reduce the number of spurious issues, that are in fact support requests, by telling people to use the mailing list at a more prominent place (README.md).

Adresses https://github.com/n1k0/casperjs/issues/1358.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,11 +64,16 @@ Run the script:
 
 ##Support
 
-Need help with getting CasperJS up and running? Got a time-consuming problem you want to get solved quickly?
+**Help request**. If you're stuck using CasperJS and don't understand how to achieve something, please [ask on the mailing-list](https://groups.google.com/forum/#!forum/casperjs) first. If discussion turns out you have detected a real issue that needs a change within CasperJS, file an issue. Please don't ask for all the kind people to write your scripts for you.
+
+**Filing issues**. It takes a lot of time to review, validate and de-duplicate filed issues. This time could be spent better on actually improving on CasperJS. Filing an issue might be a helpful contribution, but we expect you to read our [CONTRIBUTING.md](https://github.com/n1k0/casperjs/blob/master/CONTRIBUTING.md) guidelines first. 
+
+**Professional Support** Need help with getting CasperJS up and running? Got a time-consuming problem you want to get solved quickly?
 
 Get <a href="http://codersclan.net/?repo_id=32">CasperJS support on CodersClan.</a>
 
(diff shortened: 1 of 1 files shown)


istr:
Cancelled useless travis build for this PR.

README.md
```diff
@@ -64,11 +64,16 @@ Run the script:
 
 ##Support
 
-Need help with getting CasperJS up and running? Got a time-consuming problem you want to get solved quickly?
+**Help request**. If you're stuck using CasperJS and don't understand how to achieve something, please [ask on the mailing-list](https://groups.google.com/forum/#!forum/casperjs) first. If the discussion reveals that you have found a real issue that might need a change within CasperJS, file an issue. Please don't ask for all the kind people to write your scripts for you.
```
n1k0: This last sentence - which I wholeheartedly agree with - is probably unnecessary.

README.md
```diff
@@ -64,11 +64,16 @@ Run the script:
...
+
+**Filing issues**. It takes a lot of time to review, validate, and de-duplicate filed issues. This time could be spent better on actually improving on CasperJS. Filing an issue might be a helpful contribution, but we expect you to read our [CONTRIBUTING.md](https://github.com/n1k0/casperjs/blob/master/CONTRIBUTING.md) guidelines first. 
+
+**Professional Support**. Need help with getting CasperJS up and running? Got a time-consuming problem you want to get solved quickly?
 
 Get <a href="http://codersclan.net/?repo_id=32">CasperJS support on CodersClan.</a>
 
 <a href="http://codersclan.net/?repo_id=32"><img src="http://www.codersclan.net/gs_button/?repo_id=32" width="200"></a>
```
n1k0: We should probably drop this. Never had feedback at all in two years, neither from the platform not users.

n1k0:
Looks good, r+ with comments addressed.

README.md
istr: Just copied it over... :-)

README.md
istr: Ok, so drop mentioning professional support for a while until there is a solid place where it is actually offered? Or try to point to something else, e.g. https://www.bountysource.com/ ? I would be very reluctant to recommend to contact the maintainers directly.

paazmaya: Setting bounties for issues and such seems good to me

n1k0: It's mostly that this specific one hasn't got in touch for over two years now. Feels a little odd. Your call.

PR #1370, opened 2015-12-04
Fix a small typo in the doc

metamatik:
Title says it all :smile:

This looks like a simple copypasta mistake, so I just thought I would fix it quickly. I didn't believe it was necessary to follow the whole contribution pattern outlined in the repository guidelines given how trivial the change is. Please correct me if I'm wrong.

### DIFF
--- a/docs/modules/tester.rst
+++ b/docs/modules/tester.rst
@@ -920,7 +920,7 @@ Skips a given number of planned tests::
 
 **Signature:** ``tearDown([Function fn])``
 
-Defines a function which will be executed before after every test defined using `begin()`_::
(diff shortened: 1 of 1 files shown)


n1k0:
Build failure is due to an intermittent we have sometimes with phantomjs 2.0.1. Merging, thank you!

istr:
> Build failure is due to an intermittent we have sometimes with phantomjs 2.0.1.

Aha, I see. This is not intermittent, but a change that was dropped from my PR (because I could not recall why the fix I made was there: https://github.com/n1k0/casperjs/pull/1352#discussion-diff-45838428).

I will re-add these changes in a separate PR that make things work for 2.0.1 again.

Issue #1342, opened 2015-11-12
Synchronous XHR causing zero byte file downloads


laggingreflex:
Some downloads produce zero byte files. I debugged and found the root cause to be [synchronous XHR](https://github.com/n1k0/casperjs/blob/master/modules/casper.js#L243-L244). 
I wrote an async alternative and it works on the same file(s), proving indeed it's the sync xhr causing the issues. This is probably what's causing other such (zero byte files) download issues too. #73 #362 #1188

Here's a fix:

```
// this.download(downloadUrl, target); 
// ^ if this doesn't work. 
//   Try this:

this.thenEvaluate(function(url) {
    xhr = new XMLHttpRequest();
    xhr.addEventListener('load', function onTransferComplete() {
        // Set a flag when load completes, we'll keep checking it
        transferComplete = true;
    });
    xhr.open('GET', url, true); /* < ASYNC */
    xhr.responseType = 'arraybuffer';
    xhr.send(null);
}, downloadUrl);
this.waitFor(function check() {
    return this.evaluate(function() {
        return transferComplete;
    });
}, function then() {
    console.log('transferComplete!');
    var base64encoded = this.evaluate(function() {
        return btoa([].reduce.call(new Uint8Array(xhr.response), function(p, c) {
            return p + String.fromCharCode(c);
        }, ''));
    });
    /*clientutils = require('clientutils') */
    /*fs = require('fs') */
    var cu = clientutils.create(); 
    fs.write(target, cu.decode(base64encoded), 'wb');
});
```

jefleponot:
thanks @laggingreflex , very good job. Does your code work with multiple parallel downloads ?

Hi #mickaelandrieu, @paazmaya 
Could we have a casperjs "download" method with parameters to force asynchronous mode ?

istr:
When looking into the affected code, it seems that only ```clientutils.getBase64```, ```clientutils.getBinary``` ever use ```clientutils.sendAJAX``` which is causing the problem (using sync XHR).

To retain BC for 1.1 I would propose to:
- mark synchronous versions of ```clientutils.getBase64``` and ```clientutils.getBinary``` obsolete and issue a warning referencing this bug when called in sync mode
- rework ```clientutils.sendAJAX``` to support async requests / callback
- enhance ```clientutils.getBase64``` and ```clientutils.getBinary``` to be able to use a callback and work asynchronously in this case
- add casperjs.waitForAJAX to support waiting for the outcome of an asynchronous AJAX call
[...]

PR #1373, opened 2015-12-10
add features for asynchronous downloads

jefleponot:
To solve asynschronous downloads that does work with phantomjs v2 (issue :  Synchronous XHR causing zero byte file downloads #1342 ), there is a proposal PR 

for @langholz and @istr , thanks to your specifications.

PS : operation synchronously javascript is simply an absurdity because it inhibits the language spirit.
PS : PS : Could it be possible that a great tools as CasperJS be written to use asynchronous methods
(http://phantomjs.org/api/webpage/method/evaluate-async.html)
We could use it in more domains like  "chrome extension tabs" : https://developer.chrome.com/extensions/tabs#method-executeScript

Best Regards 
jef

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -386,6 +386,33 @@ Default wait timeout, for ``wait*`` family functions.
 ``Casper`` prototype
 ++++++++++++++++++++
 
+.. _casper_ajaxExists:
+
+.. index:: HTTP
+
+``ajaxExists()``
+-------------------------------------------------------------------------------
+
+**Signature:** ``ajaxExists(String|Function|RegExp test)``
+
+Checks if an ajax request has been requested by base64encode or download methods. You can pass either a function, a string or a ``RegExp`` instance to perform the test::
+
+    casper.start('http://www.google.com/', function() {
+        base64logo = this.base64encode('http://www.google.fr/images/srpr/logo3w.png');
+        if (this.ajaxExists('logo3w.png')) {
+            this.echo('Google logo loaded');
+        } else {
+            this.echo('Google logo not loaded yet', 'ERROR');
+        }
+    });
+
+    casper.run();
+
+.. note::
+
+   If you want to wait for a resource to be loaded, use the `waitForAJAX()`_ method.
+
+
 ``back()``
 -------------------------------------------------------------------------------
 
@@ -410,7 +437,7 @@ Also have a look at `forward()`_.
 ``base64encode()``
 -------------------------------------------------------------------------------
 
-**Signature:** ``base64encode(String url [, String method, Object data])``
+**Signature:** ``base64encode(String url [, String method [, Object data, Boolean async ] ] )``
 
 Encodes a resource using the base64 algorithm synchronously using
 client-side XMLHttpRequest.
@@ -443,6 +470,39 @@ encode::
 
     casper.run(function() {
         this.echo(base64contents).exit();
+    });
+
+.. note::
+
+   Use this method with waitForAJAX putting async parameter set to 'true'
+
+.. _casper_base64write:
+
+.. index:: Base64
+
+``base64write()``
+-------------------------------------------------------------------------------
+
+**Signature:** ``base64write(String str, String targetPath)``
+
+Saves a base64 string onto the filesystem after decode it.::
+
+    casper.start('http://www.google.fr/', function() {
+        var str = this.evaluate( function() {
+            function getBase64Image(img) {
+                var canvas = document.createElement("canvas");
+                canvas.width = img.width;
+                canvas.height = img.height;
+                var ctx = canvas.getContext("2d");
+                ctx.drawImage(img, 0, 0);
+                return canvas.toDataURL().replace(/[^,]*,/,"");
+            }
+            return getBase64Image(document.images[0]);
+        });
+        this.base64write(str, "logo.png');
+    });
+    casper.run(function() {
+        this.echo('Done.').exit();
     });
 
 .. index:: bypass, Step stack
@@ -699,9 +759,15 @@ Exits phantom with a logged error message and an optional exit status code::
 ``download()``
 -------------------------------------------------------------------------------
 
-**Signature:** ``download(String url, String target[, String method, Object data])``
+**Signature:** ``download(String url, String target[, String method, Object data, Boolean async])``
 
 Saves a remote resource onto the filesystem. You can optionally set the HTTP method using the ``method`` argument, and pass request arguments through the ``data`` object (see `base64encode()`_)::
+
+.. note::
+
+since phantomJS 2.0.0 synchronous downloads are forbidden : By default async is true / For backward compatibilty async is false with phantomJS 1.x.x::
+
+deprecated use::
 
     casper.start('http://www.google.fr/', function() {
         var url = 'http://www.google.fr/intl/fr/about/corporate/company/';
@@ -712,9 +778,53 @@ Saves a remote resource onto the filesystem. You can optionally set the HTTP met
         this.echo('Done.').exit();
     });
 
+better use::
+
+    var url = 'http://www.google.fr/intl/fr/about/corporate/company/';
+    casper.start('http://www.google.fr/', function() {
+        this.download(url, 'google_company.html');
+    });
+
+    casper.then(function(){
+        this.waitForDownload(url){function(){
+           this.echo('Done.');
+        });
+    });
+
+    casper.run(function() {
+        this.exit();
+    });
+
 .. note::
 
    If you have some troubles downloading files, try to :ref:`disable web security <faq_web_security>`.
+
+.. _casper_downloadexists:
+
+.. index:: download
+
+``downloadExists()``
+-------------------------------------------------------------------------------
+
+**Signature:** ``downloadExists(String|Function|RegExp test)``
+
+Checks if a download has been done. You can pass either a function, a string or a ``RegExp`` instance to perform the test::
+
+    casper.start('https://duckduckgo.com/', function() {
+        if (this.downloadExists('https://duckduckgo.com/assets/logo_homepage.normal.v107.svg')) {
+            this.echo('DuckDuckGo logo downloaded');
+        } else {
+            this.echo('DuckDuckGo logo not downloaded', 'ERROR');
+        }
+    });
+
+    casper.run();
+
+.. note::
+
+   If you want to wait for a download to be done, use the `waitForDownload()`_ method.
+
+
 
 ``each()``
 -------------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


docs/modules/casper.rst
```diff
@@ -688,10 +744,14 @@ Exits phantom with a logged error message and an optional exit status code::
...
-**Signature:** ``download(String url, String target[, String method, Object data])``
+**Signature:** ``download(String url, String target[, String method, Object data, Boolean async])``
 
 Saves a remote resource onto the filesystem. You can optionally set the HTTP method using the ``method`` argument, and pass request arguments through the ``data`` object (see `base64encode()`_)::
 
+.. note:: 
+
+since phantom 2.0.0 synchronous downloads are forbidden : By default async is true / For backward compatibilty async is false with phantom 1.x.x::
+
     casper.start('http://www.google.fr/', function() {
```
istr: I would never ever change the API behavior (i.e. default values) depending on the engine in use.

Mark this "obsolete / not recommended" in doc, throw a warning to the output when used and advise to use the newly created "waitForDownload" instead.

Better throw an error if used with an engine that does not support it than silently changing behavior here.

istr:
@jefleponot Thanks a lot for this PR :+1: .

It seems that you [have a regression with form selects in your patch](https://travis-ci.org/n1k0/casperjs/builds/96003595). Some builds fail for receipt of JSON content.

Please:
- have a look into what causes the regression(s) and fix it
- lint the code (some builds fail due to linting)

I would never silently change the API behavior due to the engine version in use, though.
Could you re-work your PR [along my proposal](https://github.com/n1k0/casperjs/issues/1342#issuecomment-162535729) and
- just keep behavior for the sync functions
- mark them obsolete / not recommended in the doc and give a hint to use the new waitFor... function s instead?
- maybe throw an error within the non-recommended functions when used with an engine that does not support them?

I would really like to integrate this PR into 1.1 because it is a major improvement.

jefleponot:
your are totaly right. You would never ever change the API behavior (i.e. default values) depending on the engine in use. The code will now throw an error.

I find my regression trouble with meld diff-tools.
I will do my best to make a PR soon.

Thanks for all your brillant analyses.

jefleponot:
partial PR done. CU next time

istr:
Nice! Thanks a lot. Restarted the failing job (known intermittent popup bug).

Will review, ask a second collaborator for opinion and then merge.

jefleponot:
I have done a lot of improvements of CasperJS for my web service supervision job, that I newer commit. if you are interested in lots of PR, it will valid them.

istr:
@jefleponot  Sounds good. Maybe it would be best to create separate feature branches to help make the different change-sets small.

Any improvement is welcome. Especially those that help to close the numerous open issues.

Thanks for your support.

docs/modules/casper.rst
```diff
@@ -688,10 +749,16 @@ Exits phantom with a logged error message and an optional exit status code::
...
 -------------------------------------------------------------------------------
 
-**Signature:** ``download(String url, String target[, String method, Object data])``
+**Signature:** ``download(String url, String target[, String method, Object data, Boolean async])``
 
 Saves a remote resource onto the filesystem. You can optionally set the HTTP method using the ``method`` argument, and pass request arguments through the ``data`` object (see `base64encode()`_)::
 
+.. note:: 
+
+since phantom 2.0.0 synchronous downloads are forbidden : By default async is true / For backward compatibilty async is false with phantom 1.x.x::
```
mickaelandrieu: PhantomJS instead please :)

modules/casper.js
```diff
@@ -254,11 +287,47 @@ Casper.prototype.back = function back() {
...
  * @param  String  data    The data to send, optional
+ * @param  Boolean async   Asynchroneous request? (default: false)
  * @return string          Base64 encoded result
  */
-Casper.prototype.base64encode = function base64encode(url, method, data) {
+Casper.prototype.base64encode = function base64encode(url, method, data, async) {
     "use strict";
-    return this.callUtils("getBase64", url, method, data);
+    async = ((typeof(async) !== "undefined")? async : false);
+	if ((phantom.version.major > 1) && (!async)) {
```
mickaelandrieu: I'd like to directly set async to true if phantomjs is > 2 instead of throw an error.

modules/casper.js
```diff
@@ -254,11 +287,47 @@ Casper.prototype.back = function back() {
...
+	   throw new CasperError("async parameter must be 'true' with base64encode");
+    }        
+    var dl = {"url": url}, dls = this.ajax.filter( function searchDownload( ajax ){
```
mickaelandrieu:
```js
var dl = {},
    dls = ...
```

modules/casper.js
```diff
@@ -254,11 +287,47 @@ Casper.prototype.back = function back() {
...
+		   return (ajax.url === url) && (typeof (ajax.filename) === "undefined");
+		});	
+	if (dls.length === 0) {
+		this.ajax.push(dl);
+	} else {
+		dl = dls[0];
+	}    
+    dl.content = this.callUtils("getBase64", url, method, data, async);
```
mickaelandrieu: hmm I don't like the use of async as variable name, this could become a reserved keyword later...

how about ``asynchronous`` instead of ``async`` ?

modules/casper.js
```diff
@@ -592,23 +661,67 @@ Casper.prototype.die = function die(message, status) {
...
  */
-Casper.prototype.download = function download(url, targetPath, method, data) {
+Casper.prototype.download = function download(url, targetPath, method, data, async) {
     "use strict";
     this.checkStarted();
-    var cu = require('clientutils').create(utils.mergeObjects({}, this.options));
-    try {
-        fs.write(targetPath, cu.decode(this.base64encode(url, method, data)), 'wb');
+    async = ((typeof(async) !== "undefined")? async : false);
+	if ((phantom.version.major > 1) && (!async)) {
```
mickaelandrieu: same here, let's limit obvious errors, and help users when we can :)

samples/download.js
```diff
@@ -7,8 +7,17 @@
 
 var casper = require("casper").create();
 
-casper.start("http://www.google.fr/", function() {
-    this.download("http://www.google.fr/images/srpr/logo3w.png", "logo.png");
+casper.on('remote.message',function(msg){
+ console.log('message : '+msg);
+// fs.write('logo4.png', msg, 'wb');
```
mickaelandrieu: remove uncommented code please

tests/suites/casper/logging.js
```diff
@@ -37,6 +37,7 @@ casper.test.begin('logging tests', 4, function(test) {
         test.assert(casper.result.log.some(function(e) {
             return e.message === 'info message' && e.level === 'info' && e.space === 'remote';
         }), 'ClientUtils.log() adds a log entry at a given level');
+console.log(this.result.log.length);
```
mickaelandrieu: why do you need to console.log here ?

mickaelandrieu:
@jefleponot great contribution ! some comments to address then. But I'm :+1: to merge this feature :)

modules/casper.js
istr: I strongly disagree, see my comments to this PR. Implicitly changing behavior due to different engine in use is evil imho. Better to explicitly tell the user that what was planned is obsolete / unsupported right away.

modules/casper.js
istr: Same here: let's tell the users "don't do this, use the way (async) that will be supported down the road". :)

istr:
@mickaelandrieu thanks for quick review. :) Maybe let's discuss the question:
- silently change API behavior based on engine in use

vs.

- warn the user she's trying to do something that won't be supported in the future

approach with @n1k0, @hexid, and @paazmaya...

jefleponot:
Another issue for you @istr : casperJS has never used "window.callPhantom" method before this PR.
It could be better to normalize this remote operation.

"window.callPhantom" is similar to the "window.postMessage" function : postMessage(data, targetDomain)

So it could be great to use the same object attributes 

The special propertys of a message event are:
<pre>
data
    The first argument of postMessage
origin
    The source domain
source
    The reference to sending window if it is possible to 
respond by calling event.source.postMessage
</pre>

chrome extensions has yet normalize communications with the "message passing" : 
https://developer.chrome.com/extensions/messaging
Message Passing is really useful to communicate between web page and "background page" (one to one) And also between many pages. It avoid to fall victim to cross-site scripting.

In the modules/casper.js file a senMessage method could be add like the <a href="https://developer.chrome.com/extensions/tabs#method-sendMessage">chrome.tabs.sendMessage </a> : sendMessage(integer tabId, any message, object options, function responseCallback)

Thanks for yours advices

mickaelandrieu:
@istr sounds OK to me :+1:

eloitay:
@jefleponot I tried patching this change into the source code myself and notice that it still produce zero byte files. So I am wondering if the way the script is written need to be changed as well in order for the bug to be fixed.

jefleponot:
Hi @eloitay 
Could you send me you code ? I could help you more

Or you can do a PR with an example.

docs/modules/casper.rst
```diff
@@ -410,7 +437,7 @@ Also have a look at `forward()`_.
 ``base64encode()``
 -------------------------------------------------------------------------------
 
-**Signature:** ``base64encode(String url [, String method, Object data])``
+**Signature:** ``base64encode(String url [, String method, Object data, Boolean async])``
```
paazmaya: This would give the impression that the third parameter is required when the second is given.
Perhaps `(String url [, String method [, Object data, Boolean async ] ] )`

jefleponot: thanks

BIGjuevos:
Do we want to keep this issue in the 1.1 release @istr @mickaelandrieu? I vote to move to a new milestone 1.2.

I feel like this would be good for a 1.2 considering our focus on squashing bugs for 1.1.  This is something I'd be happy to pick up in the future.  Feel free to assign to me.

mickaelandrieu:
Humm ok for 1.2, how many issues are blocking the release of 1.1 according to the actual milestone?

jefleponot:
test ok - PhantomJS 1.9.7 -> 2.1.1
test ok - SlimmerJS 0.8.4 -> 0.9.6

result : 6748 octets
logo

code :
````js
var casper = require("casper").create();

casper.start("http://www.google.fr/", function() {
    this.download("http://www.google.fr/images/srpr/logo3w.png", "logo.png",'GET',{},true);
});
casper.then(function(){
    this.waitForDownload("http://www.google.fr/images/srpr/logo3w.png",function(){
        console.log('download sucess');
    }, function(){
        console.log('download failed');
    });
});
casper.run();
````

It is fully functional , works in production since January 2016 with phantomJS 2.1.1 because old download method can't be used

jefleponot:
@mickaelandrieu : update done

mickaelandrieu:
@jefleponot sadly, your last merged pull request have been merged and so on you need to rebase this one :dancer: 

@casperjs/contributors so, are we agree to merge this one: my vote is :+1:

istr:
I would prefer not to break the API in this way:
- do not introduce boolean parameters
- especially do not introduce boolean parameters that change the previous "default" behaviour
- do not make the default behaviour / value dependent on the engine version in use

See https://github.com/casperjs/casperjs/issues/1342#issuecomment-207848750 what I would like to see in this PR instead:
- mark the current (broken) API functions obsolete and add a "broken" comment to the docs
- add a warning to those functions that say: "this is broken and will not work; it is obsolete and will be removed"
- create a *new* set of functions that implements what this PR achieves

jefleponot:
@istr,

Thanks for your explanation.

you are right, this PR don't suit for many reasons.
JavaScript programming is now moving towards a purely asynchronous programming. 

> 
Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check https://xhr.spec.whatwg.org/.

With PhantomJS and SlimerJS, on the opposite, methods are mixed synchronous or asynchronous : 
- open(url, callback) - **_asynchronous_**
- evaluateAsync(function, [delayMillis, arg1, arg2, ...]) - **_asynchronous_**

-  ==> page.evaluate(function, arg1, arg2, ...) - **synchronous**
- ==> page.render (filename [, {format, quality}]) - **synchronous**
- ==> page.renderBase64(format) - **synchronous**


If you look for an equivalent approach like chrome.tabs, every methods are fully asynchronous : 
- executeScript(integer tabId, object details, function **callback**)
- captureVisibleTab(integer windowId, object options, function **callback**)
(https://developer.chrome.com/extensions/tabs)

CasperJS has same trouble as PhantomJS and SlimerJS : mix synchronous and asynchronous fct.
wait(), waitFor(), waitForAlert(), waitForPopup() ...  - **_asynchronous_**
withFrame(), withPopup() - **_asynchronous_**

download() - **synchronous**
evaluate() - **synchronous**

The use of Synchronous mode, although it eases writing limits treatments.

@istr, could we propose a new set of asynchronous functions what could lead CasperJS to good javascript practices  ?

mickaelandrieu:
@jefleponot @istr I agree with Jef, and we need to keep the old api.

Could be better for now to create new functions instead of improve the existing one (and this may be a good *start* for 2.0 api :) )

sairoko12:
Someone what happened with this request?

kensoh:

Issue #1379, opened 2015-12-18
Need to forward --disk-cache-path option to PhantomJS


mepard:
--disk-cache-path is a new PhantomJS (>2.0) option to specify the location of
the disk cache. See https://github.com/ariya/phantomjs.

mickaelandrieu:
Hi,

can you link at least to a documentation link or implementation ?

Thanks

mepard:
The PhantomJS changes are at https://github.com/ariya/phantomjs/commit/c9e30ebafdeb42efe33900da7823142e146b0a1d

Just did a PR with the CasperJS changes.

--disk-cache-path=/path/to/cache/directory

PR #1380, opened 2015-12-18
Pass --disk-cache-path option to PhantomJS

mepard:
--disk-cache-path is a new PhantomJS option to specify the location of
the disk cache. See https://github.com/ariya/phantomjs.

#1379


istr:
Trivial change, checks pass. No need to add documentation for that, because [there already is an example how to pass engine options in cli.rst](https://github.com/n1k0/casperjs/blob/master/docs/cli.rst). Merging.

PR #1388, opened 2015-12-30
Fix dotnet test

### LINKED ISSUE #1385: make clitest-dotNET doesn't work as intended
Makefile:
```
clitest-dotNET:
	python tests/clitests/runtests.py mono casperjs.exe
```

runtests.py:
```
CASPER_EXEC_FILE = sys.argv[1] if (len(sys.argv) == 2) else 'casperjs'
```

Because there are two extra arguments (mono and casperjs.exe) instead of one, sys.argv has length 3 not 2 and the default casperjs is used.

I'll submit a Pull Request with the fix shortly.

mepard:
Move addition of mono from Makefile to runtests.py. Fixes #1385.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ clitest:
 	python tests/clitests/runtests.py
 
 clitest-dotNET:
-	python tests/clitests/runtests.py mono casperjs.exe
+	python tests/clitests/runtests.py casperjs.exe
 
 lint:
 	eslint .
--- a/tests/clitests/runtests.py
+++ b/tests/clitests/runtests.py
@@ -2,6 +2,7 @@
 
 import json
 import os
+import platform
 import select
 import signal
 import time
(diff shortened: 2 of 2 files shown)


istr:
Looks good; however this could actually only be tested on a windows platform (so no chance to check the difference in CI, see https://github.com/travis-ci/travis-ci/issues/216).

mepard:
I have tested it on actual Windows 10, but the mono approach seems to work, too. The CL option problem I found failed on both and is now fixed on both.

istr:
Contains the change set of https://github.com/n1k0/casperjs/pull/1387.

istr:
Looks ok; @paazmaya, @hexid, @mickaelandrieu, @n1k0 please give a second opinion and/or merge.

Merge this before https://github.com/n1k0/casperjs/pull/1389.

n1k0:
Ping @hexid; could you please have a look at this and possibly merge it if it's okay for you? Thanks a bunch.

hexid:
Looks good to me.

@mepard Thanks for the fix

PR #1404, opened 2016-01-18
Use cujos Function.bind shim; fixes #1396

### LINKED ISSUE #1396: PhantomJS #10522: Function.prototype.bind() breaks when running under 'webserver' module (test case attached)
Here's a minimal test case that will fail when run. It's using the common trick of making a short alias using Function.prototype.bind(), ie,:

    var log = console.log.bind(console)

Here's the test case:

	#!/usr/bin/env casperjs

	var casper = require('casper').create();
	var server = require('webserver').create();
	var ipAndPort = '127.0.0.1:8585';

	var log = console.log.bind(console)

	server.listen(ipAndPort, function(req, res) {

		casper.start('https://google.com');
		casper.then(function(){
			log('THIS LINE FAILS')
		});

		// This is somewhat misnamed: casper.run() starts the casper job, then runs the callback runs when complete
		casper.run(function(){
			res.close();
		});
	});
	log('CasperJS server running at http://'+ipAndPort+'/');

The `log('THIS LINE FAILS')` will fail with 

    TypeError: instanceof called on an object with an invalid prototype property

Note the object is a simple string, nothing special. It shouldn't have an invalid prototype.

istr:
This replaces the ```Function.bind``` shim taken from MDN by the cujos patch [along the lines of the associated PhantomCSS issue #10522](https://github.com/ariya/phantomjs/issues/10522#issuecomment-39248521).


istr:
@hexid, @paazmaya, @mickaelandrieu, @n1k0  Please review.

n1k0:
Looks good. Not sure about reusing the function-bind stuff, but we may want to add more polyfills in the future. r+

PR #1406, opened 2016-01-20
Mention that npm modules that require node specific features won't work

mikemaccana:
The current sentence could confuse people into thinking they can load any npm module, eg, see the confusion here:

http://stackoverflow.com/questions/24387854/use-a-node-module-from-casperjs


n1k0:
Indeed, much clearer.

n1k0:
Indeed, much clearer.

PR #1415, opened 2016-02-02
Add downloadBinary for download file correctly.

ethaizone:
I use download() to download csv and it make content gone wrong.

Ref: http://stackoverflow.com/questions/20599422/casperjs-issue-downloading-csv-file

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -445,6 +445,38 @@ encode::
         this.echo(base64contents).exit();
     });
 
+``binary()``
+-------------------------------------------------------------------------------
+
+**Signature:** ``binary(String url [, String method, Object data])``
+
+Get a resource as binary using client-side XMLHttpRequest.
+
+Example: retrieving google logo image::
+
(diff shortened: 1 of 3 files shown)


ethaizone:
If this pull request is not good, you can close it but I hope we can have better solution for it.

istr:
Basically it could be a good idea -- however test cases and updated documentation is missing.

mickaelandrieu:
@ethaizone I'm not sure we need a new method, it's better to improve the `download` function with new tests regarding the specific use cases you've met.

ethaizone:
@mickaelandrieu I think same as you but I don't know another solution so I want to tell that I found this problem too. I hope we can have better download function. I look into test case and found base64 encode is good for multi-media files too. 

I think if we can improve download(), it will better but at currently I don't have any idea.

PS. Today is first day for me to try write js testcase. Travis tell failed but it's fun. lol

ethaizone:
I test again and found this implement broke xlsx files so I will close this pull request.

Issue #1148, opened 2015-02-06
a question about casper.test.on('fail', function(){})


kantjessica:
Hi,

what does this method do? registering a new listener for "fail" event? 

what happens, if I write two casper.test.on blocks in a same test suite? If sth fails, both of the registered functions will be executed, right?

Best wishes,
Huiying

paazmaya:
It would look like an event listener for listening whenever http://docs.casperjs.org/en/latest/modules/tester.html#fail is called...

Since the `casper.test` refers to the global object, the event listener listens to all matching events.

kensoh:
@kantjessica not sure if this is still relevant to you. I think the second defined block will overwrite the first one and only the second block will be triggered. I tried testing with onStepTimeout event handling.

kensoh:
Hi @kantjessica closing this issue I believe you got the 2 answers that you are looking for.. In any case, kindly ping if you have any issues.. I'm on GitHub almost everyday these days. Or you can email at ken@tebel.org

Issue #1418, opened 2016-02-02
casper.eachThen doesn't allow falsy values.


kellyjosephprice:
When using `eachThen`, if any of the array values are falsy, they get passed in as undefined:

```
casper.eachThen([0], function (response) {
    // response.data == undefined
});
```

paazmaya:
Would you like to create a Pull Request which adds a test case for this?

kellyjosephprice:
Oops, looks like I had a typo in the PR: #1419

kensoh:
Thanks @kellyjosephprice ! I'm a CasperJS user for couple of years. Browsing through open issues now to see if anything I can help with. Saw that your PR has been merged. Closing this issue :)

PR #1419, opened 2016-02-02
refs #1148 - test undefined on step.options.data

kellyjosephprice:
Previously, `eachThen` falsy values would get munged.

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -1570,7 +1570,11 @@ Casper.prototype.runStep = function runStep(step) {
     }
     this.emit('step.start', step);
     if (this.currentResponse) {
-        this.currentResponse.data = step.options && step.options.data || null;
+        if (step.options && (undefined !== step.options.data)) {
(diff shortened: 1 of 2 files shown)


kellyjosephprice:
This should actually be referencing #1418.

modules/casper.js
```diff
@@ -1570,7 +1570,11 @@ Casper.prototype.runStep = function runStep(step) {
     }
     this.emit('step.start', step);
     if (this.currentResponse) {
-        this.currentResponse.data = step.options && step.options.data || null;
+        if (step.options && (undefined !== step.options.data)) {
```
paazmaya: How about using `typeof` check against `undefined`?

istr:
I am :+1: to merge this. @mickaelandrieu @BIGjuevos

mickaelandrieu:
Yes, why not.. this remind me we had some issues with test reporter output :/

kellyjosephprice:
Thanks all!

PR #1429, opened 2016-02-05
Hotfix: re-compiled windows binary.

istr:
I missed re-compiling the windows binary before packaging. Sorry for that.


istr:
I am merging this as a hotfix without consultation because @mepard [confirmed that the binary basically works](https://github.com/n1k0/casperjs/issues/1426#issuecomment-180097649).

mickaelandrieu:
:+1:

PR #1435, opened 2016-02-09
To more warnings with eslint !!!

jefleponot:
Hi,

I try to reduce warnings number because eslint find more than 800 warnings.

Here's a first PR to watch travis test on different browser and version

Jef

### DIFF
--- a/modules/utils.js
+++ b/modules/utils.js
@@ -28,7 +28,8 @@
  *
  */
 
-
+/*eslint func-style: [2, "declaration"]*/
+/*eslint no-use-before-define: [2, "nofunc"]*/
 var require = patchRequire(require);
 
 /**
@@ -56,9 +57,8 @@ function betterTypeOf(input) {
                 phantom.casperEngine !== "phantomjs" &&
                 '__type' in input) {
                 type = input.__type;
-            }
             // gecko returns window instead of domwindow
-            else if (type === 'window') {
+            } else if (type === 'window') {
                 return 'domwindow';
             }
             return type;
@@ -80,11 +80,11 @@ exports.betterTypeOf = betterTypeOf;
 function betterInstanceOf(input, constructor) {
     "use strict";
     /*eslint eqeqeq:0 */
-    if (typeof input == 'undefined' || input == null) {
+    if (typeof input == 'undefined' || input === null) {
       return false;
     }
     var inputToTest = input;
-    while (inputToTest != null) {
+    while (inputToTest !== null) {
       if (inputToTest == constructor.prototype) {
         return true;
       }
@@ -94,7 +94,8 @@ function betterInstanceOf(input, constructor) {
       if (typeof inputToTest == 'undefined') {
         return false;
       }
-      inputToTest = inputToTest.__proto__;
+      inputToTest = (inputToTest instanceof Object || isObject(inputToTest))
+                    && Object.getPrototypeOf(inputToTest);
     }
     return equals(input.constructor.name, constructor.name);
 }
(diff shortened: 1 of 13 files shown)


istr:
This breaks all SlimerJS tests, even after a restart.

jefleponot:
Hello @istr,

That's strange, with my ubuntu 14.4 and slimerjs 0.9.6 it works to me.

800 warnings => 74 warnings

istr:
> That's strange, with my ubuntu 14.4 and slimerjs 0.9.6 it works to me.

If you have a short look into https://github.com/n1k0/casperjs/blob/master/.travis.yml or inspect the build on travis you will see that the test suite checks for SlimerJS 0.8.4, 0.8.5, and 0.9.5, but *not* 0.9.6 (because broke coffeescript support in the test suite some time ago).

jefleponot:
I try slimerjs 0.8.5 and 0.9.5 with my ubuntu and it works. Could you check on your side not with travis ci ?

istr:
No, fails for me, too.
Using SlimerJS 0.9.5 on Ubuntu 14.04.3 and your eslint branch hangs at the same point in the test suite like in travis (betterInstanceOf is the last passing test).

```ENGINE_EXECUTABLE=~/bin/slimerjs make selftest``` is the command to run the tests.

File: modules/utils.js [hunk 1]
```diff
@@ -94,7 +94,8 @@ function betterInstanceOf(input, constructor) {
       if (typeof inputToTest == 'undefined') {
         return false;
       }
-      inputToTest = inputToTest.__proto__;
```
istr: This change breaks the SlimerJS test. This kind of regression is exactly why I proposed to
- make lint changes *only* (no functional change whatsoever)
- make *one PR per linted file*

here: https://github.com/n1k0/casperjs/issues/1430#issuecomment-180319741.

Some of the problems could be solved by adjusting the .eslintrc -- e.g. having strictly quoted keys in Objects is not a good thing to have; so I would personally still prefer
```js
        type: "assertUrlMatch",
        standard: "Current url matches the provided pattern",
        values: {
            currentUrl: currentUrl,
            pattern: pattern.toString()
```
over
```js
        "type": "assertUrlMatch",
        "standard": "Current url matches the provided pattern",
        "values": {
            "currentUrl": currentUrl,
            "pattern": pattern.toString()
```

jefleponot:
I totally agree with you. it is not a real PR but more a test to find how to optimize alerts.

jefleponot:
test passed

jefleponot:
Oups : test failed but green is displayed :-1: 
with slimerjs  we have 2 failures 

```
FAIL betterInstanceOf() detects expected constructor "undefined"
#    type: assertEquals
#    file: /home/travis/build/n1k0/casperjs/tests/suites/utils.js
#    subject: false
#    expected: true
# utils.cleanUrl() tests
```

```
FAIL Casper.getPageContent() retrieves application/json content
#    type: assertEquals
#    file: /home/travis/build/n1k0/casperjs/tests/suites/casper/content.js
#    subject: null
#    expected: "{\"dummy\":\"json\"}\n"
```

I need to investigate

istr:
> Oups : test failed but green is displayed

Good finding. This is either because the return code is not passed through by xvfb-run or because the cli tests are skipped for SlimerJS. This is one of the issues why SlimerJS support is still deemed "experimental".

istr:
> why "make" in the command ?

Because ```bin/casperjs selftest``` won't show you any version information, while ```bin/casperjs --help``` does. The make command calls both. :-)

kensoh:

Issue #1322, opened 2015-09-22
XUnit tests in Gecko (SlimerJS) defaulting to XHTML namespace + potential fix


DrCuriosity:
When generating XUnit test results using the SlimerJS engine, I'm getting an erroneous XHTML namespace attribute turning up in my test output:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites xmlns="http://www.w3.org/1999/xhtml" time="86.138">
  <testsuite name="[AC1]" tests="1" failures="0" errors="0" time="3.488" timestamp="2015-09-21T02:55:41.984Z" package="[redacted]">
    <testcase name="HTTP status code is: 200" classname="[redacted]"></testcase>
    <system-out></system-out>
  </testsuite>
</testsuites>
```

I'm using version 1.1.0-beta3, patched with the pull request from #759 to get it serializing properly.

The problem seems to be that `XMLSerializer` defaults to an XHTML namespace for `document.createElement()`. Tracing that back to ```utils.node()```, I've fixed it locally with the following:

```patch
714c714
<     var _node   = document.createElement(name);
---
>     var _node   = document.createElementNS('', name);
```

Since digging this far into the DOM is a bit arcane for me, is this the right way to go about fixing the problem? I thought about doing a ```string.replace()``` on the output of the serialization in ```xunit.js```, but that didn't feel like the right place to be doing it.

DrCuriosity:
This issue persists into 1.1.0-beta5. The fix above gets the output parsing properly in my Jenkins CI rig.

istr:
Would you mind creating a PR for the proposed bugfix? The bugfix seems trivial, but I would like to have a proper PR for
- correct attribution of authorship
- four-eyes review
- automated testing of the fix

Thank you.

DrCuriosity:
I was holding off on that until I was sure the fix was the right approach, rather than a kludge. I'll try to get a pull request sorted inside the next 24 hours.

PR #1440, opened 2016-02-14
Create DOM node with blank NS to avoid XHTML default in XUnit output

DrCuriosity:
When using SlimerJS, the XUnit output is picking up a default XHTML namespace in the `<testsuites/>` element, which can disrupt parsing of test results in CI rigs (e.g. Jenkins).

This change creates the DOM node with an explicit blank namespace in order to suppress this attribute in the serialized output.

Running CasperJS with options: `--engine=slimerjs --xunit="sample-output.xml"`

Sample output before patch:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites xmlns="http://www.w3.org/1999/xhtml" time="86.138">
  <testsuite name="[AC1]" tests="1" failures="0" errors="0" time="3.488" timestamp="2015-09-21T02:55:41.984Z" package="[redacted]">
    <testcase name="HTTP status code is: 200" classname="[redacted]"></testcase>
    <system-out></system-out>
  </testsuite>
</testsuites>
```

Sample output after patch:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites time="56.547">
  <testsuite name="[AC1]" tests="1" failures="0" errors="0" time="1.656" timestamp="2016-02-05T02:29:25.822Z" package="[redacted]">
    <testcase name="HTTP status code is: 200" classname="[redacted]"></testcase>
    <system-out></system-out>
  </testsuite>
</testsuites>
```
q.v. Issue #1322.

### DIFF
--- a/tests/suites/xunit.js
+++ b/tests/suites/xunit.js
@@ -96,3 +96,17 @@ casper.test.begin('XUnitReporter() can handle custom name attribute for a test c
     test.assertEquals(casper.getElementInfo('failure[type="footype"]').text, 'footext');
     test.done();
 });
+
+casper.test.begin('XUnitReporter() does not have default XML namespace', 1, function suite(test) {
+    var xunit = require('xunit').create();
+    var results = new tester.TestSuiteResult();
+    var suite1 = new tester.TestCaseResult({
+        name: 'foo',
+        file: '/foo'
+    });
+    results.push(suite1);
+    xunit.setResults(results);
+    casper.start().setContent(xunit.getSerializedXML());
+    test.assertNotExists('testsuites[xmlns]');
+    test.done();
+});
--- a/modules/utils.js
+++ b/modules/utils.js
@@ -726,7 +726,7 @@ exports.ms2seconds = ms2seconds;
  */
 function node(name, attributes) {
     "use strict";
-    var _node   = document.createElement(name);
+    var _node   = document.createElementNS('', name);
     for (var attrName in attributes) {
         var value = attributes[attrName];
         if (attributes.hasOwnProperty(attrName) && isString(attrName)) {


mickaelandrieu:
Hi,

thank your for pointing this issue. Let's wait for the tests build.
Maybe a specific test should/could be written to demonstrate this bug ?

Mickaël

DrCuriosity:
Test written and verified with `bin/casperjs selftest --engine=slimerjs`.

Locally, the test FAILs as expected before the patch is applied, and PASSes afterwards. Here's hoping it does the same for everyone else.

tests/suites/xunit.js
```diff
@@ -96,3 +96,17 @@ casper.test.begin('XUnitReporter() can handle custom name attribute for a test c
...
+    var xunit = require('xunit').create();
+    var results = new tester.TestSuiteResult();
+    var suite1 = new tester.TestCaseResult({
+        name: 'foo',
+        file: '/foo'
+    });
+    results.push(suite1);
+    xunit.setResults(results);
+    casper.start().setContent(xunit.getSerializedXML());
+    test.assertNotExists('testsuites[xmlns]');
```
istr: Maybe add a ```test.assertExists('testsuites');``` to point out that we expect the element, but not the attribute.

istr:
Looks good to me. Passes after restarting tests failing due to intermittent errors (known bug). I am ok to merge this. @mickaelandrieu ?

mickaelandrieu:
@istr looks good to me too, regarding your last comment.

explicit bug + explicit test to reveal it + explicit clean fix == merge (and love :D)

Thanks @DrCuriosity !

PR #1449, opened 2016-02-22
Fix windows engine flags

### LINKED ISSUE #1448: bug: no more cookies file in 1.1-beta5
Hi,
I'm working from node with the `child_process.spawn` command like follow :

```javascript
var child = spawn(
    path.join(__dirname, '../node_modules/casperjs/bin/casperjs'),
    [
        path.join(__dirname, './bots/worker.js'),
        '--tasks=' + tasks,
        '--cwd=' + this.folder,
        '--cookies-file=' + this.cookie,
        '--web-security=no',
        '--log-file=' + this.log
    ]
);
```

All filenames are absolute.
I'm using most of the flags in my `worker.js`.

Everything works fine on `1.1-beta4` but there is no cookies-file on `1.1-beta5`.
So I guess that in this latest version flags may have a problem on Windows (couldn't test it on a *nix system yet).

Tested on **Windows 7** with **node 0.10.36** and **5.5.0**, with **PhantomJS 2.0**

I'd love to help so I'll try to find out how to build the Windows executable.

yoannmoinet:
Very simple fix to be able to correctly check for native flags in the flags passed to the executable.

Before that, every flags were passed to the script, even those native.
Because `arg_match.Groups[0].Captures[0]` was holding the complete string, and not only the captured group.

It is referenced in `arg_match.Groups[1].Captures[0]` instead.

Will fix #1448.


istr:
@mickaelandrieu, @paazmaya, @hexid, @n1k0: I would like to merge this, even though it does not contain a test case.
It is a trivial fix for a flaw introduced lately.

@mepard Do you think that you could add a simple test along the lines of https://github.com/n1k0/casperjs/issues/1448#issuecomment-187331213 for that? That would be great!

mepard:
I'll take a look.

mickaelandrieu:
@istr more, we should think about appveyor asap. I dont have any experience with, but this is the only way to cover windows builds :)

mepard:
@istr I improved the options tests in #1452, which causes the Travis CI build to fail because the improved clitest-dotNET now fails. Merging both #1449 and #1452 should cause the tests to succeed.

yoannmoinet:
Would it be better to create a PR on this branch and I'll merge it?
So this one will have both.

What do you think @istr?

PR #1452, opened 2016-02-23
Detect phantomjs options not sent to phantomjs

mepard:
Check for options that were sent to the casperjs script that should have been sent to casperjs. Would have detected #1448. Validates #1449.


mepard:
The failing tests are the clitest-dotNET tests. #1449 should cause them to succeed.

istr:
Merged #1449. Please rebase this PR against current master. Thanks.

mepard:
Okay, I *think* I have it rebased. One should never tackle rebase before coffee.

Issue #1439, opened 2016-02-12
Make form optional in fillSelectors() or make method like "fillInput()"


Macek007:
Hi, 

Don't you want to make form optional in fillSelectors() or make method like "fillInput()" or setInputValue()

I found in many discussions request how to fill inputs without form, I have same issue.

I suggest to make encapsulation to method setField in casperutils.js which should do this without a lot work.

Do you want pull request?

Refactor tip: Maybe should be better to make special function for each input type (radio, select etc.) in setField method.

Thanks

P.S. CasperJS is really great!

istr:
> Do you want pull request?

Yes, every complete and useful PR that adheres to the [contributing guidelines](https://github.com/n1k0/casperjs/blob/master/CONTRIBUTING.md) is highly welcome.

Macek007:
Hi!

1000x sorry about silly question but I am newbie with GitHub pull request :(
I am ready with this feature, but I am no able to make pull request... I followed https://github.com/n1k0/casperjs/blob/master/CONTRIBUTING.md and made local copy, made branch but I am not able push it back with this error:

remote: Permission to n1k0/casperjs.git denied to Macek007.
fatal: unable to access 'https://github.com/n1k0/casperjs.git/': The requested URL returned error: 403

Did I do it in bad way? Should I made my own fork first? Thanks a lot.

mickaelandrieu:
@Macek007 it make sense because you don't have admin rights on @n1k0's repository.

You need to fork the `casperjs` repository first, using the "Fork" button on top right of Github interface when you are on the casperjs repository page.

Then you will be able to push everything, because the new created repository is yours.

When you will push new commits, Github will notify you and suggest you to make a "pull request".

Let me know if you need help :)

yoannmoinet:
@Macek007 you can follow these to understand how [PR works](https://help.github.com/articles/using-pull-requests/) and what [forking means](https://help.github.com/articles/fork-a-repo/).

kensoh:
@Macek007 I'm CasperJS user for couple of years.. Was looking through open issues and saw that you have committed code to your repo. Not sure if a pull request is sent. Bumping this up.

kensoh:
Oh looks like it has been shipped but pending committing of your PR for doc - https://github.com/casperjs/casperjs/pull/1461 :)

kensoh:
Bumping up this thread and copying maintainer @jefleponot 

[...]

PR #1459, opened 2016-02-29
refs #1439 - adding methods setFieldValue, setFieldValueName, setFiel…

Macek007:
…dValueLabel

Methods sets form field value by CSS3, XPath, name or label. Form reference is not needed.

SlimerJS ends me with errors - same as origin/master - so I ignered them.

Documentation is missing, please look first to my code.

Thanks

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -1686,6 +1686,62 @@ Casper.prototype.setContent = function setContent(content) {
     return this;
 };
 
+
+/**
+ * Sets a value to form field by CSS3, XPath selector or by it's name attribute or label text.
+ *
+ * @param String|Object selector    CSS3, XPath, name or label
+ * @param Mixed         value       Value being set
+ * @param String|Object form        (optional) CSS3 or XPath selector of form
+ * @param Object        options     Options to setFieldValue, it accepts:
+ *                                  - options.selectorType name|labes|xpath|css3 - type of selector, where
+ *                                    CSS3 and XPath(object) is autodetected (need not be set)
+ */
+Casper.prototype.setFieldValue = function setFieldValue(selector, value, form, options) {
+    "use strict";
+    this.checkStarted();
+
+    var selectorType = options && options.selectorType;
+
+    var result = this.evaluate(function _evaluate(selector, value, form, selectorType) {
+        if (selectorType) {
+            selector = __utils__.makeSelector(selector, selectorType);
+        }
+        return __utils__.setFieldValue(selector, value, {'formSelector': form});
+    }, selector, value, form, selectorType);
+
+    if (!result) {
+        throw new CasperError("Unable to set field '" + selector + " to value: " + value) +
+            ' in setFieldValue().';
+    }
+};
+
+/**
+ * Alias to setFieldValue() with implicit type name
+ *
+ * @param String        name    Name of form field
+ * @param Mixed         value   Value being set
+ * @param String|Object form    (optional) CSS3 or XPath selector of form
+ */
+Casper.prototype.setFieldValueName = function setFieldValueName(name, value, form) {
+    "use strict";
+    this.checkStarted();
+    this.setFieldValue(name, value, form, {'selectorType': 'name'});
+};
+
+/**
+ * Alias to setFieldValue() with implicit type label
+ *
+ * @param String        name    Name of form field
+ * @param Mixed         value   Value being set
+ * @param String|Object form    (optional) CSS3 or XPath selector of form
+ */
+Casper.prototype.setFieldValueLabel = function setFieldValueLabel(label, value, form) {
+    "use strict";
+    this.checkStarted();
+    this.setFieldValue(label, value, form, {'selectorType': 'label'});
+};
+
 /**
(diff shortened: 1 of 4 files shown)


mickaelandrieu:
> SlimerJS ends me with errors - same as origin/master - so I ignered them.

Seems to be good now :)

Can be useful :+1: 

Please add docs.

mickaelandrieu:
@istr my bad ! didnt saw you didnt agree with this feature and I've already merged docs ... go or no go ?

istr:
@mickaelandrieu No problem at all. I think this feature is very useful; the only thing I know @n1k0 might definitely raise an eyebrow about is the ```switch``` construct in the code. I would tend to merge it, but give me a chance for a thorough review of the code this evening.

mickaelandrieu:
:+1:  as you like ;)

modules/casper.js
```diff
@@ -1686,6 +1686,62 @@ Casper.prototype.setContent = function setContent(content) {
     return this;
 };
 
+
+/**
+ * Sets a value to form field by CSS3, XPath selector or by it's name attribute or label text.
+ *
```
istr: "...or by its name..." (its -> pronoun, it's -> it is / it has)

modules/clientutils.js
```diff
@@ -269,30 +271,13 @@
...
-                },
-                names: function(elementName, formSelector) {
-                    return this.findAll('[name="' + elementName + '"]', form);
-                },
-                xpath: function(xpath, formSelector) {
-                    return this.findAll({type: "xpath", path: xpath}, form);
-                }
-            };
-
             for (var fieldSelector in vals) {
```
istr: What is the reason to replace this ```map```able construct with a switch? (see below)

modules/clientutils.js
```diff
@@ -659,6 +644,49 @@
...
+                selector = selector.path;
+            }
+
+            switch (type) {
+                case 'css': // do nothing
+                    ret = selector;
+                    break;
+                case 'name': // convert to css 
+                case 'names': 
+                    ret = '[name="' + selector + '"]';
```
istr: If the only reason for the switch is the singular alias, you might as well add it in the object-style as seen in the existing code, i.e.

```js
var finders = {
...
labels: ...
};
finders.label = finders.labels;
```

tests/suites/casper/formfill.js
```diff
@@ -1,5 +1,6 @@
 /*eslint strict:0*/
 var fs = require('fs');
+var x = require('casper').selectXPath;
```
istr: Please try to avoid single letter variable names, even in test suites.

istr:
@mickaelandrieu Apart from some very minor questions / remarks I am ok to merge this. You decide.

mickaelandrieu:
Yes, I'm still convinced by the benefit of this contribution.

Can you address the comments @Macek007 ?

Thank you

modules/clientutils.js
Macek007: The reason was that fill() did a lot of things and I needed similar functionality for my methods - that's why I isolated of making selector (makeSelector method). 
"Finders" needed more parameters, in all cases is findAll() called etc. On other hand makeSelector() is simple method with clear interface.  
I think using switch in this case is OK but I'll would rewrite it if you want.
(btw. I think in method setField() is bad example of using mega switch with a lot of logic)

tests/suites/casper/formfill.js
Macek007: I thought that it is recommended pattern for XPath expressions according this example http://docs.casperjs.org/en/latest/selectors.html#index-2

modules/clientutils.js
istr: Totally ok to factor out the ```var finders = {...``` logic into ```makeSelector``` if you need to re-use it. But I would prefer to stick with the existing "object switch" method among other things to keep the change-set small.

tests/suites/casper/formfill.js
istr: Ok, never mind. :-) This is just a personal preference (it is easier to e.g. ```grep``` for variables with more than one letter). You decide.

mickaelandrieu:
@istr can I assign myself on it ? should be merged as the docs are already online (http://docs.casperjs.org/en/latest/modules/clientutils.html?highlight=setfieldvalue#setfieldvalue)

Should I remove the docs now, to reflect the real API ?

istr:
> @istr can I assign myself on it ? should be merged as the docs are already online

Yes, of course. Maybe just clone the PR and rebase it, if @Macek007 does not react on this ping?

jefleponot:
I can do it if you want @mickaelandrieu

mickaelandrieu:
@jefleponot go for it (thanks ! Realy sorry I'm so busy :/)

Macek007:
Hi, sorry I couldn't react earlier to this PR... it seems that @jefleponot has finished it. Do you want something else with this PR?

mickaelandrieu:
@Macek007 no, thank you for your contribution, shipped with 1.1 ;)

PR #1484, opened 2016-03-11
Fixed #1469 : Warn user when using coffee script with PhantomJS 2+

### LINKED ISSUE #1469: Newbee: Cant get CoffeeScript to work
I have casperjs 1.0.4 installed, CoffeeScript is 1.10.0. `casperjs googlelinks.js` runs fine, but `casperjs googlelinks.coffee` returns 
```
SyntaxError: Unexpected token '>'

  phantomjs://code/googlelinks.coffee:4 in injectJs
  phantomjs://code/bootstrap.js:427
```

Is this a bug or am I missing something?

mickaelandrieu:
@istr I don't know how to test it, probably using python tests but I don't know python... any hint ?

### DIFF
--- a/bin/bootstrap.js
+++ b/bin/bootstrap.js
@@ -422,6 +422,12 @@ CasperError.prototype = Object.getPrototypeOf(new Error());
 
     // casper loading status flag
     phantom.casperLoaded = true;
+    if (phantom.version.major === 2
+        && phantom.casperScript
+        && phantom.casperScript.split('.').pop() === 'coffee'
+        ) {
+        return __terminate('CoffeeScript is not supported by PhantomJS > 2.');
+    }
 
     // passed casperjs script execution
     if (phantom.casperScript && !phantom.injectJs(phantom.casperScript)) {


istr:
Maybe something like https://github.com/n1k0/casperjs/blob/master/tests/suites/casper/formfill.js#L47, but only the other way around and with a coffescript? Have to think this over, too... ;-)

mickaelandrieu:
@istr to be honest, I don't see any "elegant" workaround :/ I feel bad about this patch, and I feel bad about a specific function too: this only move the crap :/

@n1k0 what do you think ? what could be the "best" way to handle deprecations from phantomjs ?

n1k0:
I have no problem with this patch. If people want to use phantom 2 and phantom 2 doesn't support something, they shouldn't be too much surprised we don't support it either.

BIGjuevos:
I concur with @n1k0.  As long as we annotate it in the docs.

mickaelandrieu:
@BIGjuevos I add a notice in installation docs asap.

istr:
@mickaelandrieu 
> to be honest, I don't see any "elegant" workaround

No problem -- better to bail out and warn user's in an "inelegant way" rather than letting users run into unexpected trouble.  :+1: to merge this, maybe this could be replaced with an "elegant" solution later on. Life is imperfect.

mickaelandrieu:
thanks for review @istr :) I update also the docs today to inform users about this lack in phantomjs 2 and the workaround (compile into js first, then launch casperjs)

Issue #1345, opened 2015-11-16
Cannot abort request in Mac OSX


ga0:
I wrote a simple test like this:

```
//test.js
var casper = require('casper').create({
    verbose: true,
    logLevel: 'debug',
    pageSettings: {
        loadImages: false
    }
});
casper.on('resource.requested', function(req, networkRequest) {
    this.echo("> " + req.url);
    networkRequest.abort();
});
casper.on('resource.received', function(resp) {
    this.echo("< " + JSON.stringify(resp));
});

var url = casper.cli.get('url');
casper.start(url);
casper.run();
```

and run it:

```
casperjs test.js --url=https://github.com --ssl-protocol=any
```

**The 'networkRequest.abort()' doesn't work in macosx, the request has been sent as normal.**

casper version: 1.1.0-beta3
phantomjs version: 1.9.7

This example runs perfectly in my windows pc and centos7 server.

Acostan:
Same issue here: running phantomjs: 2.0.0 casper 1.1.0-beta3

istr:
To address this we would need two things:
- migrate the tests that check ```request.abort``` from ```tests/suites/casper/request.coffee``` to plain JS to be future proof and work with PhantomJS 2.0.0 and later
- add OS/X tests to our builds https://docs.travis-ci.com/user/osx-ci-environment/

It would help a lot if you could do any of these in a branch and create a PR for that.

As a start, could you please check if the current test suite passes the coffee / abort tests locally with your setup (phantom up to 1.9.8 only)? (```make test``` in a checkout of current master)

istr:
The test case you presented fails for me, too (and honestly I don't believe it could work correctly on the other platforms you mentioned). This is because you used the wrong event names to hook in.

If you have a look at https://github.com/n1k0/casperjs/blob/master/tests/suites/casper/request.js#L14 you will see that you need to hook into ```page.resource.requested``` instead of ```resource.requested```.

The following test works for  me (Linux / PhantomJS 1.9.7 and 2.0.0).

```js
//test.js
var casper = require('casper').create({
    verbose: true,
    logLevel: 'debug',
    pageSettings: {
        loadImages: false
    }
});
casper.on('page.resource.requested', function(req, networkRequest) {
    this.echo("> " + req.url);
    networkRequest.abort();
});
casper.on('page.resource.received', function(resp) {
    this.echo("< " + JSON.stringify(resp));
});

var url = casper.cli.get('url');
casper.start(url);
[...]

PR #1490, opened 2016-03-18
refs #1345 - fixed by improving clean method

BIGjuevos:
It was changed so that the clean method is more apt to its name.  It will clean any popups that are not longer valid.

I feel that this is the best way to implement this operation.  I'm pretty sure this won't break BC.

### DIFF
--- a/modules/pagestack.js
+++ b/modules/pagestack.js
@@ -50,22 +50,20 @@ exports.Stack = Stack;
 Stack.prototype = [];
 
 /**
- * Cleans the stack from closed popup.
+ * Cleans the stack from any closed popups.
  *
- * @param  WebPage  closed  Closed popup page instance
  * @return Number           New stack length
  */
-Stack.prototype.clean = function clean(closed) {
+Stack.prototype.clean = function clean() {
     "use strict";
-    var closedIndex = -1;
+    var self = this;
+
     this.forEach(function(popup, index) {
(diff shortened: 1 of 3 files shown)


BIGjuevos:
I tested this in phantomjs 1.9.18 and phantomjs 2.1.1.

modules/pagestack.js
```diff
@@ -50,22 +50,20 @@ exports.Stack = Stack;
 Stack.prototype = [];
 
 /**
- * Cleans the stack from closed popup.
+ * Cleans the stack from any closed popups
  *
```
istr: . Fullstop. ;-)

istr:
As far as I can see the ```clean``` method is not part of the public / documented API anyway. So :+1: to merge. Ping @mickaelandrieu for 2nd opinion.

BIGjuevos:
First CI build succeeded, 2nd one failed after fixing a typo in the comments.  The build hung.  It can safely be ignored.

mickaelandrieu:
:+1v thanks @BIGjuevos and Jef for the solution :)

mickaelandrieu:
Also, would you mind to squash your commits?

BIGjuevos:
@mickaelandrieu squashed.  Sorry about that.

mickaelandrieu:
thanks ! :)

Issue #1488, opened 2016-03-18
0 as a timeout value in wait() method


wskorodecki:
`CasperError: wait() only accepts a positive integer > 0 as a timeout value`

I think it is better to allow 0 as a value, because sometimes it is very useful to disable all application delays in configuration by changing some variable to 0 instead of adding if-statements for this purpose or adding 1 as a value which is ugly.

In other words - if the timeout value is omitted or 0 the method should interpret this as "call the function immediately". Just like the setTimeout() function works in JavaScript:

`setTimeout(function() {console.log('It works!');}, 0);`

istr:
A PR to resolve this issue is highly welcome.

kensoh:
Hi @wskorodecki this seems to be possible now, I can do a wait with 0 as parameter.

kensoh:
Thanks @wskorodecki! Looks like your pull request is already merged, closing this issue for now. Probably no change in documentation needed (existing doc looks ok) but copying @jefleponot for his call.

PR #1493, opened 2016-03-21
refs #1488 - improvements in wait* methods

wskorodecki:
Allowing 0 as a timeout value.
Added missing checks.
Issue: #1488

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -2040,6 +2040,36 @@ Casper.prototype.warn = function warn(message) {
 };
 
 /**
+ * Helper functions needed in wait*() methods. Casts timeout argument to integer and checks if next step
+ * function is really a function and if it has been given (if required - depending on isThenRequired flag).
+ *
+ * @param   Number   timeout        The max amount of time to wait, in milliseconds
+ * @param   Function then           Next step to process (optional or required, depending on isThenRequired flag)
+ * @param   String   methodName     Name of the method, inside of which the helper has been called
+ * @param   Number   defaultTimeout The default max amount of time to wait, in milliseconds (optional)
+ * @param   Boolean  isThenRequired Determines if the next step function should be considered as required
+ * @returns Number
+ */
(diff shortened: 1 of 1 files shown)


modules/casper.js
```diff
@@ -2185,6 +2192,13 @@ Casper.prototype.waitForAlert = function(then, onTimeout, timeout) {
...
 Casper.prototype.waitForPopup = function waitForPopup(urlPattern, then, onTimeout, timeout) {
     "use strict";
+    if (then && !utils.isFunction(then)) {
+        throw new CasperError("waitForPopup() next step definition must be a function");
+    }
+    timeout = ~~timeout;
+    if (timeout < 0) {
+        throw new CasperError("waitForPopup() only accepts an integer >= 0 as a timeout value");
+    }
     return this.waitFor(function() {
```
istr: Please put this block of code in a helper function and call it rather than copying it around.

istr:
Very useful. :+1: to merge this. As this is basically internal (refactoring) only, I do not see the need to update tests and/or documentation here. @mickaelandrieu @BIGjuevos

modules/casper.js
```diff
@@ -2040,6 +2040,36 @@ Casper.prototype.warn = function warn(message) {
...
+        var isFunction = utils.isFunction(then); // Optimization to perform "isFunction" check only once.
+
+        if (isThenRequired && !isFunction) {
+            throw new CasperError(methodName + "() needs a step function");
+        } else if (then && !isFunction) {
+            throw new CasperError(methodName + "() next step definition must be a function");
+        }
+    }
+
+    timeout = ~~timeout || ~~defaultTimeout;
```
mickaelandrieu: shouldn't be ``var timeout = ~~timeout || ~~defaultTimeout;`` ?

mickaelandrieu:
Ok to me too :) thank you @wskorodecki  and @istr for review

PR #1465, opened 2016-03-02
Fix encoded urls when using phantomjs 2.1.0 or newer

mepard:
I'm not intending for this to land, yet. It demonstrates #1464.

If you swap the order of the urls, the test will pass.

PR #1417, opened 2016-02-02
BugFix and Evol MouseEvents

jefleponot:
Hi @istr , @mickaelandrieu , others

When we use Native Events something doesn't work. In fact we can't use these events on selector when there aren't in the visible area (inside the page viewport and scrollPosition areas). It doesn't work also if scroll position is not fit page correctly. So it is not user-friendly ... because you can't maximize viewport without having some decrease in reactivity (SlimerJs).

"ContextMenu" doesn't work except onto the main html element.

With this PR you can control more finely event on canvas, image ...

PR #1495, opened 2016-03-21
Fix encoded urls when using phantomjs 2.1.0 or newer

### LINKED ISSUE #1464: Encoded urls fail when using phantomjs 2.2.1
If the first url(s) opened requires encoding, the response lacks the status. After a url is opened that does not require encoding, urls that require encoding work okay. In the following sequence, url 1 fails, url 2 works, url 3 works:

    page%20_1.html
    page_2.html
    page%20_3.html

In this sequence, all work okay:

    page_1.html
    page%20_2.html
    page%20_3.html

Stay tuned for a PR with a test that demonstrates the problem.

mepard:
Fixes #1464. Rebased #1465.

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -1209,7 +1209,9 @@ Casper.prototype.handleReceivedResource = function(resource) {
         return;
     }
     this.resources.push(resource);
-    if (utils.decodeUrl(resource.url) !== this.requestUrl) {
+
+    var checkUrl = utils.ltVersion(phantom.version, '2.1.0') ? utils.decodeUrl(resource.url) : resource.url;
+    if (checkUrl !== this.requestUrl) {
         return;
     }
     this.currentHTTPStatus = null;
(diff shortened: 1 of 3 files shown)


modules/casper.js
```diff
@@ -2637,7 +2639,8 @@ function createPage(casper) {
     };
     page.onResourceRequested = function onResourceRequested(requestData, request) {
         casper.emit('resource.requested', requestData, request);
-        if (requestData.url === casper.requestUrl) {
+    	var checkUrl = utils.ltVersion(phantom.version, '2.1.0') ? utils.decodeUrl(requestData.url) : requestData.url;
```
istr: whitespace / lint

istr:
@mickaelandrieu @BIGjuevos I would like to merge this. Please have a look. Thank you.

modules/casper.js
```diff
@@ -1209,7 +1209,9 @@ Casper.prototype.handleReceivedResource = function(resource) {
         return;
     }
     this.resources.push(resource);
-    if (utils.decodeUrl(resource.url) !== this.requestUrl) {
+
+    var checkUrl = utils.ltVersion(phantom.version, '2.1.0') ? utils.decodeUrl(resource.url) : resource.url;
```
BIGjuevos: Hoist your variables?

BIGjuevos:
No objections to the code overall, just some minor style updates and it get's my :+1:

Thanks for your hard work @mepard

istr:
We should keep this in mind and maybe put that code snippet in a helper function. This pattern occurs in other places, see e.g. #1417.

mickaelandrieu:
:+1:  ok too :)

regarding your last comment @istr, I agree, let's see how this pattern evolve and we will refactor when required, what do you think ?

istr:
@mickaelandrieu 
> regarding your last comment @istr, I agree, let's see how this pattern evolve and we will refactor when required

Absolutely. Let us land this now and refactor when necessary.

istr:
Merging because it fixes a real bug and has a very useful test-case. Postpone linting / refactoring.

PR #1506, opened 2016-03-31
move to archived versions of phantomjs since bitbucket throttles

BIGjuevos:
Let's try pulling in phantomjs from somewhere else, and see what happens.  This requires some minor tweaks to the travis yml to like gz files as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,17 @@ language: generic
 
 before_install:
   - echo "Installing $ENGINE $ENGINE_VERSION from $ENGINE_ARCHIVE_URL"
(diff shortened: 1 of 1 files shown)


BIGjuevos:
@istr @mickaelandrieu Builds are now pulling from a github release.  Also updated travis config to support both bz2 and gz compress tar files for future use.

mickaelandrieu:
Sounds good to me :) thank you !

BIGjuevos:
@istr any objections to merging this?  I think this should go in before anything else so we can stabilize the builds.

istr:
Ok, thank you. Merging this now (sorry, but always try to work from past to current).

PR #1082, opened 2014-11-05
SlimerJS has some options that start with - instead of -- and CasperJS w...

DarrenCook:
(see mailing list thread "How to pass params to the engine?")

PR #1507, opened 2016-04-01
refs #1082 - putting existing PR on top of master

BIGjuevos:
SlimerJS has some options that start with - instead of -- and CasperJS will now forward them with this commit.

This is blatant theft from #1082 with just a rebase on top of the current master.


mickaelandrieu:
:+1:

istr:
Thanks for the rebase. Merging in favor of  identical, outdated #1082.

Issue #1504, opened 2016-03-30
CasperJS throwing error when run with PHP


Sahilgoyal190:
Casper Js is throwing an error when run with PHP exec() command on windows environment, Although its running file if use directly on command line (cmd)
##

`_ 'phantomjs' is not recognized as an `internal` or external command, operable program or batch file._`

istr:
Duplicate of #1355 and #1366. Using PHP as a proxy / web service for CasperJS does not seem to be a good idea, anyway and I don't think that we will try to support it in the near future.

BIGjuevos:
Not suggesting we re-open this issue, but to at least point @Sahilgoyal190 in the right direction.

PHP tends to not have the same environment variables when it is executing than when you do as a user.  Make sure your paths are right in your PHP configs, and if you have any further issues, the fine folks over at the PHP Stack Overflow will be able to point you in the right direction.

istr:
@BIGjuevos maybe drop a hint somewhere in the documentation. Duplicate issues are more often than not a hint that something is at least expected to work, but not well documented why it can't.

BIGjuevos:
I've created an example of how to create a web service with casper/phantomjs by blatantly stealing @mickaelandrieu's code and putting it into a [new section in the docs](http://docs.casperjs.org/en/latest/cookbook.html#creating-a-web-service).

Issue #1366, opened 2015-12-02
CasperJS Doesn't output to stdout


optikalefx:
This is a really strange issue, but I've gotten the issue located down to casper. I have a PHP script that runs `shell_exec` which just runs a casper script.  But NULL is returned. If I run a phantomjs script, I get the output just fine.

However if I run that same casper script from terminal, it does output. So something in casper is swallowing the output.  PhantomJS outputs using console.log and phantom.exit. Casper never outputs to shell_exec.

CasperJS version 1.1.0-beta3 at /root/casperjs, using phantomjs version 1.9.8

istr:
Please try to change the call to ```exec``` (http://php.net/manual/en/function.exec.php) and analyze / report the exit code of the call. With ```shell_exec``` you can't tell if the call to casper failed or did not produce output (http://php.net/manual/de/function.shell-exec.php).

Probably your call from PHP to CasperJS plainly fails (I would suspect issues with the ```PATH```).

This is a typical question that should be directed to [the mailing list first](http://groups.google.com/group/casperjs) because it might just be a problem with your specific PHP / CasperJS setup and not an issue with CasperJS itself.

optikalefx:
For those coming here in the future. The solution ended up being to avoid using PHP to execute the call. So the setup for me is

- PHP app server >  
- cURL with information (not the actual request) to a server on the local network so the main server isn't doing the scraping. > 
- nodeJS which exec's a child process to run casper > 
- capserJS script > 
- which runs PhantomJS

istr:
@optikalefx Thanks for reporting back.

There is a [related issue wanting to do the same](https://github.com/n1k0/casperjs/issues/1355).

It would be great if you could provide your node.js script that addresses your use-case so we could add it as an example.

istr:
Ok, closing. Thanks for submitting your script. If you would like to go through the effort to create a ready-made example and create a PR that would be highly welcome. If you don't and we integrate an example based on your script we would mention your authorship and add a reference to your comment containing it.

optikalefx:
@istr I would imagine it would be most helpful starting from node instead of mine which starts from PHP?

soulteary:
hi, @optikalefx , you can use `system` and others cmd to excute casperjs ,eg: https://github.com/Pantimos/Dashboard.Pantimos/blob/master/eve-include/mock.class.php

BIGjuevos:
[...]

Issue #1355, opened 2015-11-23
Add example how to run CasperJS as a web service


timwis:
I'm trying to expose my CasperJS test as a REST API. I [posted](http://stackoverflow.com/questions/33863608/how-can-i-expose-my-casperjs-tests-as-their-own-rest-api/33864717) this question on stack overflow and was directed to [this answer](http://stackoverflow.com/questions/15852987/casperjs-passing-data-back-to-php/16489950#16489950), which suggested I use PhantomJS' built-in `webserver`. That seems to work, but there are two unexpected behaviors:

1. If you just use `casper.run()` without a callback function, casper exits after the first HTTP request, killing the server
2. If you provide a callback function to `casper.run()` (as is done in the example) but don't call `this.exit()` within it, the server stays up but, as the comment below that answer suggests, memory accumulates.

Is this intended behavior? Is there a better approach?

istr:
It seems that @optikalefx [has found a solution for this](https://github.com/n1k0/casperjs/issues/1366#issuecomment-161662959).

I would appreciate it if this could be disclosed, so we can add it as an example.

optikalefx:
Sure thing, here is the full node script. I run this with `pm2 start proxy.js`. The reason I have the __ properties is because I pass data from a cURL script to hit this URL.  I do that so my scraping is done on a different server than the main server. So that script (in PHP) does a cURL hitting this page (@ port 3000) with data about what URL to scrape and what data to use.

```
/* eslint-env node */
/*eslint no-console: 0*/
/*eslint quotes: 0*/
/*eslint strict: 0*/

var app = require('express')();
var bodyParser = require('body-parser');
var exec = require('child_process').exec;
var fs = require('fs');
var shellescape = require('shell-escape');

app.use(bodyParser.json());
app.post('/', function (req, res) {

	var url = req.body.__url;
	var action = req.body.__action;
	var header = req.body.__header;
	var creds = req.body.__creds;
	delete req.body.__url;
	delete req.body.__action;
	delete req.body.__header;
	delete req.body.__creds;

	creds = JSON.parse(creds);
	creds = JSON.stringify(creds);

	var postData = JSON.stringify(req.body);

	var args = [
		'casperjs',
		'scrape.js',
		'--web-security=no',
		'--ignore-ssl-errors=true',
		'--ssl-protocol=any',
		url,
		action,
		creds,
		postData
	];

	var cmd = shellescape(args);

	exec(cmd, function (error, stdout) {
[...]

PR #1508, opened 2016-04-01
Start of cookbook in the documentation

BIGjuevos:
This addresses the common questions found within issues like #1355, #1366, and #1504.  It also introduces the concept of a cookbook to the documentation so future common patterns can be easily documented.

@istr @mickaelandrieu @n1k0


mickaelandrieu:
Good idea :)

PR #1509, opened 2016-04-03
introduce 'known issues' to documentation

BIGjuevos:
We should let people know in the documentation that we are aware of certain problems, and if available what the fix is.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -24,6 +24,7 @@ CasperJS_ is a navigation scripting & testing utility for the PhantomJS_ (WebKit
    faq
    changelog
    upgrading/index
+   known_issues
    credits
    license
 
--- a/docs/known_issues.rst
+++ b/docs/known_issues.rst
@@ -1,0 +1,24 @@
+.. _known_issues:
+
+.. index:: Known Issues
+
+Known Issues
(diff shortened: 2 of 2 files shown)


mickaelandrieu:
:+1:  what do you think  @istr ? I fear a possible duplicate with faq section ?

BIGjuevos:
Is it worth putting this as a sub section under the FAQ docs?  Or just converting this into FAQ style questions?

Though, putting it under FAQ feels like shoving a square peg into a round hole to me.

mickaelandrieu:
Finaly, thinking about it make me feel it's a bad idea to put it in FAQ section.
Can you add a link to faq in the installation section ?

istr:
> I fear a possible duplicate with faq section ?

IMHO duplication would hurt less than omission. I am not sure, though, if it is worth duplicating what is already tracked in github tickets. Maybe we stick with the policy to document only "known issues" there that we deem largely unresolvable (i.e. we cannot solve by means of a bug-fix or work-around), like the one @BIGjuevos started with.

tl;dr -- :+1: to merge this.

mickaelandrieu:
thanks @BIGjuevos

Issue #1497, opened 2016-03-22
Document regressions since phantomJS 2


mickaelandrieu:
- [x] support of coffee (wontfix, document workarounds ?)
- [x] multiple file upload regression (since phantom 2.0.1 => https://github.com/ariya/phantomjs/issues/12506)

BIGjuevos:
These have been documented with #1510.  Closing issue.

Issue #1203, opened 2015-04-02
Keep Support For Coffeescript With PhantomJS 2.0


timscott:
To my horror, PhantomJS 1.9 stopped working for me on a certain website. To my delight, PhantomJS 2.0 fixes it. But then to my horror again, apparently PhantomJS has dropped support for Coffeescript.

https://github.com/ariya/phantomjs/issues/12410

It seems my choices are:

 1. Add add a compile step (with cake, I guess) on deployment. In development, manually compile after every code change (ouch!).
 2. Convert my thousands of lines of existing Coffeescript to JS, and leave Coffeescript behind (double ouch!).

I guess there is a third choice: somehow compile at runtime. I guess I can find a way to do this in my app before calling CasperJS. However, I wonder if there is way to add this step to CasperJS. It would invoke whenever the file is Coffescript and PhantomJS is >= 2.0.

Is that is possible? Is it desired? If desired, where would the code go? I'm guessing somewhere around [here](https://github.com/n1k0/casperjs/blob/master/bin/bootstrap.js#L127)?

I would be willing to take this on if it's desired. I must solve it for myself asap.

petetnt:
For workaround number 1 you could use something like [grunt-contrib-coffee](https://github.com/gruntjs/grunt-contrib-coffee) with [grunt-contrib-watch](https://github.com/gruntjs/grunt-contrib-watch) to automatically compile the .coffee files after every code change.

timscott:
I'm running CasperJS from a Ruby on Rails app. It turns out to be pretty easy because the ```coffee-rails``` gem gives me ```CoffeeScript.compile``` (via the ```ruby-coffee-script``` gem).

I created a little runtime build system that compiles the script and all possible dependencies just before it's executed. It build only once in production and every time in other environments.

It solves my issue nicely, but it would be even nicer if CasperJS would handle it. The most valuable coding work is done with the delete key!

paazmaya:
It is less burden for a project such as CasperJS to reduce the amount of dependencies, here namely CoffeeScript, in similar way as PhantomJS did.

Would @timscott be willing to create a PR for documentation against the `phantomjs-2` branch, where he could share how his setup was done?

BIGjuevos:
These have been documented with #1510.  Closing issue.

@timscott We would love it if you could share your findings in the [new cookbook](http://docs.casperjs.org/en/latest/cookbook.html), or even just drop us a link to the gist so we can integrate it.

[...]

PR #1510, opened 2016-04-04
refs #1497,#1203 - add two more known issues

BIGjuevos:
This addresses the request to document two other known issues as mentioned in #1497


mickaelandrieu:
@BIGjuevos would you mind to add a note in installation section about coffeescript support ? :)

BIGjuevos:
QuickStart or Installation?

mickaelandrieu:
Maybe after the pre requisites => http://docs.casperjs.org/en/latest/installation.html#prerequisites

BIGjuevos:
@mickaelandrieu After looking in the docs, I decided to update it in quick start as well as introduction.  Since I think both are valid points of entry and one or the other may be skipped.

mickaelandrieu:
thank you @BIGjuevos !

Issue #777, opened 2013-12-20
casperjs download file character encoding utf8 issue


hsnelamir:
I'm trying to download the following csv file using casperjs

``` js
var casper = require('casper').create({});
casper.start('http://69.50.252.196/');
casper.then(function(){
    this.download('http://69.50.252.196/download.csv', 'downloaded.csv');
});
casper.run(function() {
    this.exit();
});
```

download.csv

```
إختبار  Arabic
آزمایشی Persian
测试  Chinese
測試  Chinese
испытание   Russian
परीक्षा Hindi
δοκιμή  Greek, Modern (1453-)
테스트 Korean
טעסט    Yiddish
テスト Japanese
பரிட்சை Tamil
```

but the downloaded file always corrupted and got corrupted characters

downloaded.csv

```
"%.*('1 Arabic" 
"2E'ج4ج Persian
Kص  Chinese
,f  Chinese
8A?KB0=85   Russian
*0@M7> Hindi
´؟؛¹¼®  Greek"   Modern (1453-)"
"L¤¸    Korean" 
"طâلط   Yiddish"    
"ئ¹ب    Japanese"   
"ھ°؟ںحڑب    Tamil"  
```

```
casperjs --version
1.1.0-beta3
phantomjs --version
1.9.2
```

rpacheco-aa:
Has this been merged?

paazmaya:
Still seems to be the "old" https://github.com/n1k0/casperjs/blob/master/modules/casper.js#L596

@yujiroarai and @r8k seem to have something done, but would there be anything for a PR (with tests included)?

yujiroarai:
@paazmaya what is old ?

paazmaya:
The current code:
```
fs.write(targetPath, cu.decode(this.base64encode(url, method, data)), 'wb');
```
as discussed above

qknight:
so no fix yet? this is really broken as it is now. please fix ;-)

mickaelandrieu:
@qknight a "workaround"  is provided,  but feel free to make a pull request :)

bolonio:
Have you tried to run the script with the output encoding command line parameter?

casperjs --output-encoding=cp866 [params] [filename]

celine-m-s:
+1
I'm having big troubles with the encoding. Are you planning to merge the PR? Thanks!!

jefleponot:
Hi @celine-m-s 
Which PR are you talking about ?

celine-m-s:
[...]

PR #1519, opened 2016-04-11
Added tests of encode() functions

mickaelandrieu:
With this sample, I dont meet any issue ... let's travis check all PhantomJS versions :)


mickaelandrieu:
@istr hmmm my tests pass ... how to fix  #777 ? :-)

istr:
@mickaelandrieu Strange, indeed. But the test is nice; I would like to merge this. @BIGjuevos any idea why this test does not reveal the problem in #777?

BIGjuevos:
Is it the combination of the non-english and english characters adjacent to each other?

It could also be a false positive if you copy-pasted from github into the test and the CSV from the same source rather than natively generating the characters from within code.  Character encoding is strange in my experience, and tends to behave when copy pasting.

mickaelandrieu:
> It could also be a false positive if you copy-pasted from github into the test and the CSV from the same source rather than natively generating the characters from within code. 

I did :p any idea on how I can generate this kind of characters within code ?

BIGjuevos:
> I did :p any idea on how I can generate this kind of characters within code ?

On mac, go into keyboard preferences and add another keyboard layout that is a troublesome layout, and mash some keys to get it directly inputted.

tests/suites/casper/encode.js
```diff
@@ -12,6 +12,29 @@ casper.test.begin('base64encode() and download() tests', 2, function(test) {
...
+
+        var stream = fs.open('__base64encode.csv', 'r'),
+        expectedValues = [
+            'إختبار,',
+            'آزمایشی,',
+            '测试',
+            'испытание',
+            'परीकi'
+        ],
+        i = 0,
```
mickaelandrieu: as a note for myself: should be `var i = 0;`

jefleponot:
it is quite strange to me to add this kind of test but code seems ok.
agree to merge

mickaelandrieu:
@jefleponot we try to cover some edge errors but the test don't cover all the cases, according to @BIGjuevos comments :/

mickaelandrieu:
Note I have never reproduced this issue, so it's hard for me to write an efficient test :)

jefleponot:
@mickaelandrieu , are you ok to merge this PR ?

mickaelandrieu:
I can't validate my own pull requests :cat:

jefleponot:
lol, I'm ok for this PR. I just want to know if you still want that I add your PR.

mickaelandrieu:
Yes ! :+1:

PR #1536, opened 2016-04-14
add BC for coffeeScript for PhantomJS >= 2.

jefleponot:
@Istr, 

Here is a solution to keed the backward compatibility for coffeeScript.
This Feature adds the coffeeScript support for phantomJS >= 2 (exactly all browser that not contain in require extensions 'coffee')

Now you have the choice : 
- to keep coffeeScript support
- to stop coffeeScript support
- to limit coffeeScript to old browser version.

Best Regards, jef
personally I had never developped coffeescript code ... perhaps one day or not :fr:


BIGjuevos:
Is it worth copy-pasting the coffee-script.js?

Can we not include this as a npm dependency?

jefleponot:
lol, as you wish @BIGjuevos - this js file is minify so I had to ignore eslint scheduling

could you help me ? I don't use node.js and npm dependency

mickaelandrieu:
Any chance to see this as an extension of core, and installable using npm ? I want to keep CasperJs as simple as possible.

jefleponot:
Lol @mickaelandrieu ,
it's really hard to find a solution to backward compatibility without making concessions.

mickaelandrieu:
@jefleponot the thing is I don't see why all CasperJS users should "load" this extension since phantomJS 2 dont support it natively. With the rise of ES6, is Coffeescript still relevant ?

More, this could be worth to investigate on a way to plug extensions to CasperJs instead of complete the api: CasperJS is already too big, and should stay a thin layer on top of PḧantomJS.

I do agree with the old dream of @n1k0 to remove the tester api and made it optional: if we can do it, we can probably plug another testing frameworks, another reporters, another communities and made the CasperJS community grown :)

Tell me what do you think, and if you have some ideas on how to do it "smoothly".

PS: I appreciate a lot your contributions, I also try to made this softaware the easier to maintain for all of us :)

jefleponot:
@mickaelandrieu , I don't understand why Ariya Hidayat (PhantomJS) and Laurent Jouanneau (SlimerJS) have chosen to support coffeeScripts at first (as Nicolas Perriault in CasperJS).
As they decided to stop using coffeeScripts, I understand now that many users are disappointed and very frustrated.
My programs are only written in javascript so I am not affected by this decision.

So as we talked previously of BC with Ingo Stuck @istr , I just write 6 lines of code :paperclip: to show everyone how keeping coffeescript cost. 
3 proposals : 
- write a doc to explain how to do without coffeescript (and clean code)
- write the doc how to hack casperJS to use coffeescript
- provide coffeescript support instead of browsers

PS: I appreciate a lot yours advices and all your work,  I think software is easy to maintain if functions are small and well documented.

My dream to test web site : a free user friendly interface like https://ghostinspector.com/ (no more program for testers)

n1k0:
> as Nicolas Perriault in CasperJS

Because at the time ES2015 didn't exist, and CoffeeScript was a much welcome improvement over ES5. Now ES2015 is out and rather widely supported (starting with Babel), promoting CS in any way doesn't make much sense to me.

istr:
> promoting CS in any way doesn't make much sense to me.

I agree. Let us just drop coffeescript support for the 2.0 target.

mickaelandrieu:
@istr by the way we should keep this contribution somewhere for all who need to keep CoffeeScript. Perfect usecase for a cookbook article, right ?

BIGjuevos:
Unless @jefleponot wants to, I'll put together a PR for the cookbook.

jefleponot:
Hi Everybody,

> Unless @jefleponot wants to, I'll put together a PR for the cookbook.

nice idea.

In my opinion, Backward Compatibility is important but in that case, Coffee-script launching must be outsourced.
So it could be good to add a single hook in bootstrap.js for support JavaScript compilers extensions (online or direct) .*Not only for coffee-script developers*

Just have a look to : 
https://github.com/jashkenas/coffeescript/wiki/List-of-languages-that-compile-to-JS
there are more than 380 ways to generate or interpret JavaScript.

Regards

PR #1537, opened 2016-04-14
Change all of the URLs due to the migration

BIGjuevos:
I think i found them all

@mickaelandrieu @istr


mickaelandrieu:
:+1:  thank you @BIGjuevos

PR #1538, opened 2016-04-14
update the .org links as well

BIGjuevos:
another obvious quick update that we should do to prevent confusion.


mickaelandrieu:
Thank you @BIGjuevos

PR #1539, opened 2016-04-15
fixes #1535 - make xml exporting idempotent, and actually render

### LINKED ISSUE #1535: first study casperjs,run my demo,but  xunit xml has no testcase,why?
```js
casper.test.begin('Study', 1, function suite(test) {
    casper.start("https://github.com/n1k0/casperjs", function() {
        test.assertTitle("GitHub - n1k0/casperjs: Navigation scripting & testing utility for PhantomJS and SlimerJS", "verify title");
	});
    casper.run(function() {
        test.done();
    });
});
```

runcommond: casperjs test demo.js --xunit=log.xml
Test file: demo.js
# Study
PASS Study (1 test)
PASS verify title
PASS 1 test executed in 9.038s, 1 passed, 0 failed, 0 dubious, 0 skipped.

but log.xml content is:
<?xml version="1.0" encoding="UTF-8"?><!DOCTYPE casperjs PUBLIC "-//CasperJS//XUnit Test Results" "testsuites"><testsuites/>

I used phantomjs 1.9.8
Trouble to help me, I was a rookie.

BIGjuevos:
fixes regression introduced by 3742e14

### DIFF
--- a/modules/xunit.js
+++ b/modules/xunit.js
@@ -81,14 +81,11 @@ exports.create = function create() {
 function XUnitExporter() {
     "use strict";
 
-    // Since we are outputting XML, let's do our own document type
-    var documentType = document.implementation.createDocumentType("casperjs", "-//CasperJS//XUnit Test Results", "testsuites");
+    this.setupDocument();
 
-    this._xmlDocument = document.implementation.createDocument("", "", documentType);
-    this._xml = this._xmlDocument.appendChild(this._xmlDocument.createElement("testsuites"));
-
-    // Initialize everything else
+    // Initialize state
     this.results = undefined;
+    this.rendered = false;
 }
 exports.XUnitExporter = XUnitExporter;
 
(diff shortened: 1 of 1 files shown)


istr:
It seems that none of our tests check that ```--xunit``` actually includes the test results in the output.

@BIGjuevos could you add a simple test to the client util test script ```./tests/clitests/runtests.py``` ?

BIGjuevos:
Yeah. I'll add a check.

mickaelandrieu:
We probably need to setup a jenkins to check ? I'm very busy theses days :/ 
About this contribution, this is better cause we use getXML() function again :)

istr:
> We probably need to setup a jenkins to check?

No, we just need
- a simple test case with e.g. two suites (one passes, one fails)
- a reference output.xml that includes the outcome of these
- a diff / expect for the clientutil suites

> I'm very busy theses days :/

Me too... * sigh *

istr:
I manually checked this regression fix. Let us publish this now and hand in a cli test case later. Lots of people seem to rely on the xunit output.

PR #1548, opened 2016-04-20
update newPage()

entrptaher:
Add version alert in newPage() function.


istr:
@entrptaher Thank you. Merging without 2nd opinion because documentation-only and obviously OK.

PR #1555, opened 2016-04-22
closes #548

### LINKED ISSUE #548: onLoadError callback, load.failed event not working as expected/documentation very thin
while trying to find a way to count the number of >400 <599 http status responses, i noticed that onLoadError and load.failed are not behaving like i expected after reading the documentation.

this event:
```javascript
/* Emitted when PhantomJS’ WebPage.onLoadFinished event callback has been called and failed.*/
casper.on('load.failed', function errorhandler() {
    casper.echo('load.faild');
});
```
Is not fired if http://mypage.com/index.html contains images that generate a 404
is not fired if http://mypage.com/doesnotexit.html is 404
is only fired if http://doesnotexist (non-existing domain) is faild

```javascript
/*A function to be called when a requested resource cannot be loaded*/
casper.options.onLoadError = function (casper, resource) {
    console.echo("loaderror");
    console.log('received: ' + resource.url + ' ' + resource.status);
};
```
this is, in the 3 cases stated above, never be fired. same for onLoadFinished, no idea what these events are for.

```javascript
/*Emitted when any resource has been received.*/
casper.on('resource.received', function(resource) {
    casper.echo(resource.url + ' ' + resource.status);
});
```
this also fires if the resource is 404, or the request is completely failed (like http://doesnotexist). for me as external user, the resource is not received, so this event should not fire.

even if my complains above are invalid and are the expected result, the events need to be more documented. it is not clear when they are fired. for each resource-related event, make clear if it related to a main resource (page browsed to) or for subresources (i.e. images on that page) or both. and also how resource errors work (404 does not seem to be an error for casperjs/phantomjs).

BIGjuevos:
Adds a note in the cookbook about how to build something like this.

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -75,3 +75,116 @@ You can then access the results via an HTTP POST request:
   curl --data "casperjs" http://127.0.0.1:8585/
 
 The above command would search for "casperjs" on google and return a JSON array of results.  This is a trivial example and can be expanded into something more complex.
+
+Script to automatically check a page for 404 and 500 errors
+-----------------------------------------------------------
+
+.. code-block:: javascript
+
+  var casper = require("casper").create({
+    pageSettings: {
+      loadImages: false,
+      loadPlugins: false
+    }
+  });
+  var checked = [];
+  var currentLink = 0;
+  var fs = require('fs');
+  var upTo = ~~casper.cli.get('max-depth') || 100;
+  var url = casper.cli.get(0);
+  var baseUrl = url;
+  var links = [url];
+  var utils = require('utils');
+  var f = utils.format;
+
+  function absPath(url, base) {
+    return new URI(url).resolve(new URI(base)).toString();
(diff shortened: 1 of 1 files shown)


istr:
imho it is better not to reference gists but to copy the gist into https://github.com/casperjs/casperjs/tree/master/samples and reference that instead.

gists tend to disappear at some point in the future leaving broken references that may stay unnoticed for a long time.

BIGjuevos:
@istr, that't funny.  I had the exact same opposite thought.  "Why copy paste it when we have a link to a really nice gist with comments."

I'll pull it in as well, and reference the gist since it has lots of comments.

BIGjuevos:
@istr - Updated.

istr:
> I'll pull it in as well, and reference the gist since it has lots of comments.

:+1: I am angry at broken gist links very often, so I'd always prefer to keep a copy. Merging.

Issue #1560, opened 2016-04-28
Please make __utils__.sendAJAX support custom headers


Sayalic0:
That's very useful, and every time I need it, I will just modify the source code of `__utils__.sendAJAX` to support custom headers. 

There is a [StackOverflow post](http://stackoverflow.com/questions/24268647/sending-http-headers-in-form-submission-casperjs) which is about the custom header feature, and one can see it takes dozens lines of code to implement such a feature.

Why I don't just post a pull request? Because I'm a beginner for JavaScript. I think my implementation is dirty, so I post a issue here.

paazmaya:
Would you mind sharing your implementation, even if you feel it is "dirty" ?

Sayalic0:
You may see Pull Request #1562

Sayalic0:
Travis-CI build failed for just 1 test case, but I think my code has nothing to do with `Casper.base64encode()`. I read the test code, it's just retrieve a local image file and encode it to base64 string. But in my code, it returns a zero length string instead of 6410 as expected. I have no idea to fix that, so could you please give me some advices?

Failed test logs:
```
FAIL Casper.base64encode() can retrieve base64 contents
#    type: assertEquals
#    file: /home/travis/build/casperjs/casperjs/tests/suites/casper/encode.js:9
#    code: test.assertEquals(image.length, 6160, 'Casper.base64encode() can retrieve base64 contents');
#    subject: 0
#    expected: 6160
```

kensoh:
Hi @Sayalic0 bumping this up :)

kensoh:
Hi @Sayalic0 I saw that the reference implementation was made, but was it by another user improving on your original implementation? Hmmm...

Want to thank you for your contribution!! Also copying current maintainer Jef @jefleponot for info. Closing this issue for now as there are no outstanding actions I think. But do give a shout if anything!

PR #1562, opened 2016-04-30
add sendAjax with custom headers

Sayalic0:
Adding sendAjax with custom headers. See [Issue#1560](https://github.com/casperjs/casperjs/issues/1560).

### DIFF
--- a/modules/clientutils.js
+++ b/modules/clientutils.js
@@ -849,11 +849,12 @@
          * @param   String   url      Url.
          * @param   String   method   HTTP method (default: GET).
          * @param   Object   data     Request parameters.
+         * @param   Object   headers  Request headers.
          * @param   Boolean  async    Asynchroneous request? (default: false)
          * @param   Object   settings Other settings when perform the ajax request
          * @return  String            Response text.
          */
-        this.sendAJAX = function sendAJAX(url, method, data, async, settings) {
+        this.sendAJAX = function sendAJAX(url, method, data, headers, async, settings) {
             var xhr = new XMLHttpRequest(),
                 dataString = "",
                 dataList = [];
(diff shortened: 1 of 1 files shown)


paazmaya:
In order to be backwards compatible, check that if the `headers` argument is boolean, then it would in fact be the `async`, etc

istr:
There is already a `settings` argument. No need to change the signature, just pass headers as sub-object / property of the `settings` and check for them.

modules/clientutils.js
```diff
@@ -849,11 +849,12 @@
...
          * @param   String   method   HTTP method (default: GET).
          * @param   Object   data     Request parameters.
+         * @param   Object   headers  Request headers.
          * @param   Boolean  async    Asynchroneous request? (default: false)
          * @param   Object   settings Other settings when perform the ajax request
          * @return  String            Response text.
          */
-        this.sendAJAX = function sendAJAX(url, method, data, async, settings) {
+        this.sendAJAX = function sendAJAX(url, method, data, headers, async, settings) {
             var xhr = new XMLHttpRequest(),
```
istr: Do not change the signature, pass it within settings.

modules/clientutils.js
```diff
@@ -864,6 +865,9 @@
             if (settings && settings.overrideMimeType) {
                 xhr.overrideMimeType(settings.overrideMimeType);
             }
+            for(var header in headers) {
```
istr:
```js
if (settings && settings.headers) {
  for (var header in settings.headers) {
...
```

istr:
@Sayalic0 This PR is a good start, even if you qualify yourself as "a beginner for JavaScript." (#1560). Just integrate the supposed change, add a test case and documentation and this PR is very likely to be accepted.

Thanks for your contribution. :-)

Sayalic0:
Thanks for kind reply of you guys. I will check your advice carefully and refine my PR.

Sayalic0:
@istr I refined my code and wrote docs about the my new feature. Now I get stuck in writing test. I just have no idea about how to write a proper unit test for `sendAJAX`. Should I start a web server in a unit test, then send a request to that web server and check the header?

istr:
@Sayalic0 Thanks for your effort. Please be patient, I do not have the time to give a good advice before next weekend. Thanks for understanding.

Sayalic0:
@istr No hurry. Take your time and thanks for you reply.

PR #1563, opened 2016-04-30
Add real life example of waitWhileVisible()

entrptaher:


BIGjuevos:
Simple docs update.  Merging.

Issue #1566, opened 2016-05-02
'click' documentation issue


ianchanning:
I'm using casper js v1.1.1 with phantom js v2.1.1 on Linux mint.

**Edit: I've found the documentation now and submitted pull request https://github.com/casperjs/casperjs/issues/1567 for this now. I would still suggest that the casperjs-docs repo be removed or changed to point at the correct docs.**

There is a problem with the ['click' sphinx documentation](http://docs.casperjs.org/en/latest/modules/casper.html#click). I'm adding it here as there doesn't appear to be issues that I can add to https://github.com/casperjs/casperjs-docs.

This is connected to #1532 as it is one of the main examples given on the home page which should be a self contained script that you can copy and paste which works.

Currently the click documentation has the following example:

``` js
casper.start('http://google.fr/');

casper.thenEvaluate(function(term) {
    document.querySelector('input[name="q"]').setAttribute('value', term);
    document.querySelector('form[name="f"]').submit();
}, 'CasperJS');

casper.then(function() {
    // Click on 1st result link
    this.click('h3.r a');
});

casper.then(function() {
    // Click on 1st result link
    this.click('h3.r a',10,10);
});

casper.then(function() {
    // Click on 1st result link
    this.click('h3.r a',"50%","50%");
});


casper.then(function() {
    console.log('clicked ok, new location is ' + this.getCurrentUrl());
});

casper.run();
```

The problem with that is that you can't copy and paste it in to a file and run it straight off.

First you need `var casper = require('casper').create();`, which is easy enough if you've read other examples.

However then it just produces no output and no errors and it's not clear why. 

I tracked it down to the fact that the first `this.click('h3.r a');` is successful but then the later clicks aren't on the assumed google page and so crash, preventing any further output - so the last `console.log` never gets run.

Bizarrely the 'latest' docs for ['click' in github](https://github.com/casperjs/casperjs-docs/blob/master/modules/casper.rst#click), would actually work, it just looks like someone has added two extra examples of how to use click that break the code.

``` js
casper.start('http://google.fr/');

casper.thenEvaluate(function(term) {
    document.querySelector('input[name="q"]').setAttribute('value', term);
    document.querySelector('form[name="f"]').submit();
}, { term: 'CasperJS' });

casper.then(function() {
    // Click on 1st result link
    this.click('h3.r a');
});

casper.then(function() {
    console.log('clicked ok, new location is ' + this.getCurrentUrl());
});

casper.run();
```
[...]

PR #1567, opened 2016-05-02
Update casper.rst

ianchanning:
removed two click examples, so that the script will run without crashing once you add in `var casper = require('casper').create();`

Based on https://github.com/casperjs/casperjs/issues/1566 that I raised. This click example runs correctly and gives an example of the two extra parameters.

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -496,19 +496,8 @@ Example::
 
     casper.then(function() {
         // Click on 1st result link
-        this.click('h3.r a');
-    });
-
-    casper.then(function() {
-        // Click on 1st result link
-        this.click('h3.r a',10,10);
-    });
-
-    casper.then(function() {
-        // Click on 1st result link
         this.click('h3.r a',"50%","50%");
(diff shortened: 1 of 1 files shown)


BIGjuevos:
If we are going to take away some samples that are incorrect, can we add some in that are correct instead?

mickaelandrieu:
@BIGjuevos yup we shouldn't remove samples that do not work anymore but fix them. 

:-1: for me

istr:
Agree with @mickaelandrieu and @BIGjuevos.

ianchanning:
@BIGjuevos @mickaelandrieu I looked at fixing the script with all the examples but you have to add in manual `back` events, then it becomes a large unwieldy script that is hard to understand. 

It's then showing three ways of doing exactly the same thing with slightly different arguments. 

The example I chose used all the arguments of the method so that they can be seen in action. But at the same time it's easy to understand what it is doing.

Surely to help first time users you don't want to show them all the possibilities, just the best simple example that they can run?

PR #1569, opened 2016-05-03
Add a drag&drop test example

ebrehault:


paazmaya:
Could you include the complete working code in the example?

ebrehault:
Do you mean the code located in the codepen page?

istr:
I think the sample is clear, even without the html that it tests. It would be necessary for a test-case, but for the cookbook it sufficiently demonstrates the case (imho). @BIGjuevos @mickaelandrieu ?

BIGjuevos:
It makes sense to me even without the accompanying HTML.

I'll let @mickaelandrieu voice his opinion and merge.

mickaelandrieu:
I've suggested him to share some cookbooks,  @ebrehault is author of an ebook about testing using CasperJS. 

:+1:

Ps: you don't need my voice to decide,  I mean I trust the double eyes strategy of @istr :) 

Ps2: my students have discovered CasperJS today...  and they like it <3

mickaelandrieu:
BTW thanks @ebrehault for sharing,  don't hesitate to contribute again, we love contributions!

ebrehault:
@mickaelandrieu I planned to push more examples but there are broken (they target public sites like Gmail or Yahoo! Weather, and the markup changed a lot since I wrote those examples).

Issue #1532, opened 2016-04-13
Add Better Samples to Help Users Getting Started.


istr:
As a recent hot discussion and a look into the open issues show, we should think about better samples and documentation to get users started.

While there are scores of good samples, code snippets, and even testing modules building on top of CasperJS out in the wild, we see lots of false issues filed against CasperJS (basically all those tagged "Help Request") and some [frustrated users turning their back on CasperJS before they even got started](https://github.com/n1k0/casperjs/issues/1530). From my own experience I remember that the "learning curve" to get a grip on CasperJS is not as flat as it could be.

Being totally aware that CasperJS is by no means a "point-and-click" solution, maybe we should
- try to add some sort of primer with basic "DOs" and "DON'Ts"
- add some more complex examples that work on / with popular JS frameworks (e.g. angular, react to name a few)
- try to optimize the guidance / initial path through the existing documentation

Ping @mickaelandrieu, @BIGjuevos, @jefleponot for thoughts and input.

BIGjuevos:
I agree.  I can take on some backbone and angular example apps.

- I think the lowest hanging fruit is the Quick Start and optimizing it.
- Is it worth updating the casperjs.org homepage to encourage people to go through an improved quick start?

istr:
> worth updating the casperjs.org homepage

Definitely. It could help to give a better guidance using 
- less links
- a few simple steps, e.g.
  1. download
  2. quickstart
  3. samples / primer
- improve on the weight (and position in the page) the links have according to relevance for newcomers

mickaelandrieu:
Agree too.  I will suggest a new article on "how can you debug a CasperJS script?"

BIGjuevos:
So, as you can tell by the timestamp, I got some inspiration to give the homepage a revamp and gave it a quick go.  Let me know what you think @istr @mickaelandrieu 

First draft, what do you guys think?

![](http://i.imgur.com/ctKXU8i.png)

istr:
Looks good... I would remove the ```-g``` flag from the npm install command though, because you can do a quick test without super-user rights with a local installation.

ianchanning:
Is there a reason why the documentation typically leaves out the first line:

```javascript
var casper = require('casper').create();
```

There are quite a few scripts that are almost self contained, i.e. they have 

```javascript
casper.start('...');
...
casper.run();
```

But then anyone learning can't just copy and paste and run that script they always have to add in the first line.

[...]

PR #1585, opened 2016-05-14
Add a function to clean Google search result urls.

entrptaher:


istr:
Nice. :+1: to merge this. ping @BIGjuevos @mickaelandrieu

BIGjuevos:
It was [suggested](https://github.com/casperjs/casperjs/issues/1532#issuecomment-217737064) that we not even crawl the almighty Google for this front page example.

However, this is a marked improvement over the current.  Merging.

@entrptaher If you would like to rewrite the section so that it demonstrates the same feature without encouraging people to scrape Google, it would be really appreciated.

Issue #178, opened 2012-07-11
Casper.page.content adds html tags for non-html content types


n1k0:
We should add a `getPageContent()` method to casper to access raw body contents when response content-type is _not_ `text/html`.

tobeorla:
Why this issue is closed? This bug remains unfixed and "view-source:" hack doesn't work.

dotmechanize:
This issue does not appear to be fixed. Any updates?

When requesting pure JSON from a known good (serves up only JSON) URL I also get the HTML tags mentioned above.

istr:
This is fixed in master, see [the test proving it](https://github.com/n1k0/casperjs/blob/master/tests/suites/casper/content.js).

NOTE however, that you need your server to deliver the correct ```Content-Type``` header (```application/json```) -- there is no content sniffing whatsoever.

dotmechanize:
Thanks for the feedback. I'll have to give it a try.

For what it's worth, I discovered that the problem was intermittent. In cases where the returned JSON was not 'prettified' all worked well. If I set my server's JSON output to be 'pretty' then the HTML tags appeared and caused JSON.parse() to fail.  My workaround has been to simply strip out all HTML with a regex before attempting to create a JSON object. This approach is working for me.

entrptaher:
This is not fixed yet.
Only if somehow this is implemented, that'd be great.

```js
JSON.parse(this.getPageContent().replace(/<\/?[^>]+(>|$)/g, ""));
```

istr:
@entrptaher Could you explain what you mean with your code snippet and/or a test case that demonstrates this still does not work? Did you verify that your server returns the correct `Content-Type` header? (see above)

istr:
@dotmechanize 
> If I set my server's JSON output to be 'pretty' then the HTML tags appeared and caused JSON.parse() to fail.

Are you sure that setting the "prettify" option did not change the response's `Content-Type` header?

entrptaher:
# CODE 1(Works as Expected)
```js
var casper = require('casper')
    .create();

casper.start();
casper.open('http://localhost/');
casper.echo("...");
casper.then(function dumpHeaders() {
    this.currentResponse.headers.forEach(function (header) {
        console.log(header.name + ': ' + header.value);
    });
});
casper.then(function getthecontents() {
    console.log(this.getPageContent());
});
casper.run(function () {
    this.exit();
});
```
##Output: ```
...
X-Powered-By: Express
[...]

PR #1587, opened 2016-05-14
Creates getPlainText()

entrptaher:
Creates a very simple function to return plain data that is currently visible to the viewer.
https://github.com/casperjs/casperjs/issues/178#issuecomment-219220406

Should the naming of the function matter?

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -980,6 +980,17 @@ Casper.prototype.getPageContent = function getPageContent() {
 };
 
 /**
+ * Retrieves current page contents in plain text.
+ *
+ * @return String
+ */
+Casper.prototype.getPlainText = function getPlainText() {
+    "use strict";
+    this.checkStarted();
+    return this.page.framePlainText;
+};
+
+/**
  * Retrieves current document url.
  *
  * @return String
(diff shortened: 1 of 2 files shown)


istr:
Yes, nice... However, I would prefer a test that fetches the plain text from the test server, preferably without the `Content-Type` header set. This way we could actually test what went wrong in #178.

So in `tests/selftest.js`:

Add a bogus `notype` entry to `contentTypes` that matches the file extension `.notype` and set the `type` to null:

```js
var contentTypes = {
    html: {type: 'text/html', binMode: false},
    js: {type: 'application/javascript', binMode: false},
    json: {type: 'application/json', binMode: false},
    txt: {type: 'text/plain', binMode: false},
    png: {type: 'image/png', binMode: true},
    notype: {type: null, binMode: false},
    _default: {type: 'application/octet-stream', binMode: true}
};
```
only add the header if its type is set:
```js
if (contentType.type) {
    headers['Content-Type'] = contentType.type;
}
```

Then add a file `tests/site/plaintext.notype` with the contents to be checked, fetch this in your test case and see if you get the raw text contents like expected.

This would cover all the weirdness discussed in #178.

entrptaher:
Ah, I closed this by mistake.

PR #1602, opened 2016-05-26
removed unavailable usage option from CLI usage

mickaelandrieu:
Hi,
i've looked in history of project and I think this option was never been available or tested or even documented.

As I dont know what was the expected or planned behavior, I suggest to remove it for now.

Mickaël


istr:
I am ok to merge this. @BIGjuevos ?

PR #1610, opened 2016-06-02
can't open program #1180

ideayuye:
This issue is because node can't current parse the command width absolute path have space.Like this

     node  D:\\Program Files\\nodejs\\node_modules\\casperjs\\bin\\casperjs\\bootstrap.js

execute this command will take an error "can't open D:\\program".Node parse this as "node  D:\\Program".

So we can add en enclosing double quotation marks to this command .

    node  "D:\\Program Files\\nodejs\\node_modules\\casperjs\\bin\\casperjs\\bootstrap.js"

It will execute correctly.This issue maybe not only exist in win.Since win default install casperjs to "C:\Program Files",the win's user the more likely  to encounter the problem.

With my modify ,execute "capserjs selftest", will 3 failed.There are two  of them is the test merge string is not rigorous.Another one ,I don't find the reason.

It seems that this problem has been dealt with before #1319.

### DIFF
--- a/bin/casperjs
+++ b/bin/casperjs
@@ -130,11 +130,13 @@ for arg in SYS_ARGS:
 CASPER_COMMAND = [ENGINE_EXECUTABLE]
 CASPER_COMMAND.extend(ENGINE_ARGS)
 CASPER_COMMAND.extend([
-    os.path.join(CASPER_PATH, 'bin', 'bootstrap.js'),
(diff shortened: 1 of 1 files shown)


istr:
Why did you close this one? Would you mind re-opening and rebasing against current master?

ideayuye:
Because "The Travis CI build failed". Perhaps because there are three test cases did not pass.

PR #1642, opened 2016-07-10
restore a previous PR for coffee backward compatibility

jefleponot:
restore a previous PR for coffee backward compatibility from #1246

### DIFF
--- a/modules/utils.js
+++ b/modules/utils.js
@@ -467,7 +467,7 @@ function isJsFile(file) {
     var valid = Object.keys(require.extensions).map(function(val) {
         return val.replace(/^\./, '');
     }).filter(function(ext) {
-        return ext === 'js';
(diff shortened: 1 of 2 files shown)


modules/utils.js
```diff
@@ -467,7 +467,7 @@ function isJsFile(file) {
     var valid = Object.keys(require.extensions).map(function(val) {
         return val.replace(/^\./, '');
     }).filter(function(ext) {
-        return ext === 'js';
+        return ext !== 'json';
```
BIGjuevos: If we are going to do this, I'd rather be explicit.
`'js' || 'coffee'`

jefleponot: done

mickaelandrieu:
Thank you @jefleponot  (sorry for delay)

PR #1647, opened 2016-07-14
Downloaded.error emit

TheGrandPackard:
In writing a program that will download files, I found it very useful to listen for the downloaded.file event, but had no way to test for the download failing. I added a new custom event for an error so that i can listen to it.

I also added an entry for the docs so that other users can use it.

### DIFF
--- a/docs/events-filters.rst
+++ b/docs/events-filters.rst
@@ -156,6 +156,13 @@ Emitted when the ``Casper.die()`` method has been called.
 
 Emitted when a file has been downloaded by :ref:`Casper.download() <casper_download>`; ``target`` will contain the path to the downloaded file.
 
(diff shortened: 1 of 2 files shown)


mickaelandrieu:
Hi, good idea!

Ok for me

BIGjuevos:
I like this idea too.

If it passes, I'll merge it in.

mickaelandrieu:
Wow ! 2 maintainers active at same time oO :D

BIGjuevos:
Fastest PR EVAR!

TheGrandPackard:
Woot! This is a really great suite for testing (and automating some really pesky web forms). Thanks for taking time to look at the PR and get it done!

PR #1653, opened 2016-07-20
add runScript method : equivalent to Webpage#evaluateAsync

jefleponot:
Hi,

CasperJS has never coded the PhantomJS method "evaluateAsync". But as @mickaelandrieu said on #1373 : 

> it could be better for now to create new functions instead of improve the existing one

because of JavaScript programming is now moving towards a purely asynchronous programming.

So this PR allows to run asynchronous evaluate with a "casper.runScript" method.
I don't named it "evaluateAsync" because (it's ugly) it is not the same prototype as the PhantomJS method.

casper.runScript accepts a callback function with one argument : the return data or the "callCasper" data.

samples : 

``` js
casper.runScript(function(text) {
    return text + document.title;
}, 'new title: ', function(ret) {
    this.echo(ret);
});


casper.runScript(function(text) {
    document.addEventListener('click',function(){
        callCasper(document.title);
    });
   return "ignored";
}, function(title) {
    this.echo(title);
});
```

return : 

> new title: CasperJS, a navigation scripting and testing utility for PhantomJS and SlimerJS
> CasperJS, a navigation scripting and testing utility for PhantomJS and SlimerJS

Regards
jef

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -1593,6 +1593,42 @@ Casper suite **will run**::
 
 Binding a callback to ``complete.error`` will trigger when the ``onComplete`` callback fails.
 
+``runScript()``
+-------------------------------------------------------------------------------
+
+**Signature:** ``runScript(Function fn[, arg1[, arg2[, …]]], callback)``
+
+Basically `PhantomJS' WebPage#evaluateAsync <http://phantomjs.org/api/webpage/method/evaluateAsync.html>`_ equivalent. Asynchronous evaluates an expression **in the current page DOM context**::
+
+    casper.runScript(function(text) {
+        return document.title;
+    }, 'new title : ',
+    function(ret) {
+        this.echo(ret);
+    });
+
+    casper.runScript(function(text) {
+        callCasper(document.title);
+    }, function(ret) {
+        this.echo(ret);
+    });
+
+
+    casper.runScript(function(text) {
+        document.addEventListener('click',function(){
+            callCasper(document.title);
+        });
+    }, 
+    function(ret) {
+        this.echo(ret);
+    });
+.. note::
+
+   For forcing return use the "callCasper" method
+
+.. topic:: Understanding ``evaluate()``
+
+
 .. index:: Scroll
 
 ``scrollTo()``
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -143,6 +143,7 @@ var Casper = function Casper(options) {
     this.mouse = mouse.create(this);
     this.popups = pagestack.create();
     // properties
+    this.callbacks = {};
     this.checker = null;
     this.currentResponse = {};
     this.currentUrl = 'about:blank';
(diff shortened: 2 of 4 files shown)


jefleponot:
@mickaelandrieu 

any suggestion about this PR ?

mickaelandrieu:
I'm in favor of it.

Sadly I no longer use casperjs, so I'm not the good person to decide on this I guess :/

Mickaël

BIGjuevos:
If the operation is going to be async.  The code should be self-documenting and include `async` or `withCallback` somewhere in the function name.

jefleponot:
Hi @BIGjuevos ,

I concur that function name must be self evident :
prototype function **xxx** (function evaluate,args...object, function callback)

Have you a proposal ?

For example, here is a list of names used by other asynchronous libraries :
**executeScript** <= [chrome webextensions ](https://developer.chrome.com/extensions/tabs#method-executeScript)
**runJavaScript** <= [Qt 5] (http://doc.qt.io/qt-5/qml-qtwebengine-webengineview.html#userScripts-prop)
**executeAsync** <= [Webdriver.io ](http://webdriver.io/api/protocol/executeAsync.html)

I can't understand that you mean with including  "withCallback" .

jefleponot:
what do you prefer 1 or 2?

1. using callCasper
```js
casper.runScript(function(text) {
    document.addEventListener('click',function(){
        callCasper(document.title);
    });
   return "ignored";
}, function(title) {
    this.echo(title);
});
```
2. using function passed in parameter
```js
casper.runScript(function(text, done) {
    document.addEventListener('click',function(){
        done(document.title);
    });
   return "ignored";
}, function(title) {
    this.echo(title);
});
```
it is also possible to mix 1 and 2.

jefleponot:
Hi @mickaelandrieu , @istr , @BIGjuevos 

Another way to proceed,

keep all the casper functions and only work on the "then" function.
Have a sample : 

Now (synchronous mode)

```js
var title = casper.evaluate(function() {
    return document.title;
});
```

Suggestion ( Asynchronous mode ) :

```js
casper.evaluate(function() {
    return document.title;
}).then( function(title){
  casper.echo(title);
});
```

----------------------------------
another sample 


Now

```js
casper.on('remote.callback', function(title){
   casper.echo(title);
});

casper.evaluate(function() {
    setTimeout(function(){
        window.callPhantom(document.title);
   }, 3000}
});
```

Suggestion :

```js
casper.evaluate(function() {
    setTimeout(function(){
    callCasper(document.title);
   ,3000}
}).then( function(title){
  casper.echo(title);
});
```

I think it could be a coherent way to improve CasperJS while maintaining its simplicity.

to compare CasperJS method to Webdriver commands
CasperJS is easiest 83 methods vs 196 webdriver commands  :smile:

jefleponot:
@BIGjuevos, could I propose a PR for one of this evolution ?

@istr , any idea ?

kensoh:

PR #1657, opened 2016-07-21
fix slimer.version

jefleponot:
utils.ltVersion(slimer.version, '0.10.0')

instead of

utils.ltVersion(phantom.version, '0.10.0')

### DIFF
--- a/modules/casper.js
+++ b/modules/casper.js
@@ -1229,7 +1229,7 @@ Casper.prototype.handleReceivedResource = function(resource) {
     this.resources.push(resource);
 
     var checkUrl = ((phantom.casperEngine === 'phantomjs' && utils.ltVersion(phantom.version, '2.1.0')) ||
-                   (phantom.casperEngine === 'slimerjs' && utils.ltVersion(phantom.version, '0.10.0'))) ? utils.decodeUrl(resource.url) : resource.url;
+                   (phantom.casperEngine === 'slimerjs' && utils.ltVersion(slimer.version, '0.10.0'))) ? utils.decodeUrl(resource.url) : resource.url;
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
Hum, I'm not sure that `slimer` variable exists the only engine variable is ``phantom`` and you should refers to it.

I think, we should rename ``phantom`` to ``engine`` but it's a lot of works for no big gain.

jefleponot:
you are wrong.

https://docs.slimerjs.org/current/api/slimer.html

with slimerJS the variable phantom.version contains the version of PhantomJS to which SlimerJS is compatible (read-only)

So ((phantom.casperEngine === '**slimerjs**' &&  utils.ltVersion(**phantom.version, '0.10.0'**))
is totaly no sense

mickaelandrieu:
Make sense then, but what is the value of `phantom.version` when slimerjs is used as the engine ?

jefleponot:
slimerjs-0.8.4 => phantom.version equivalent to {"major":1,"minor":9,"patch":0}
slimerjs-0.8.5 => phantom.version equivalent to {"major":1,"minor":9,"patch":0}
slimerjs-0.9.6 => phantom.version equivalent to {"major":1,"minor":9,"patch":2}
slimerjs-0.10.0 => phantom.version equivalent to {"major":1,"minor":9,"patch":8}

mickaelandrieu:
o_O wow.

:+1:  to me, thank you @jefleponot for hint this bug !

ping @casperjs/contributors for merge please :)

Mickaël

BIGjuevos:
2 legit 2 quit :+1:

Issue #1677, opened 2016-09-04
Failed to search the svg node using XPath


sishen:
I have the svg icons in the page and want to get the node using xpath selector.

```
this.click(x("//*[@id='container']/div[2]/div[8]/svg:svg"));
```

It raises the exception and reported that it failed to get the node in the document.

jefleponot:
Good PR

do you mind to test others namespaces ?

```js
function nsResolver(prefix) {
  var ns = {
    'xhtml' : 'http://www.w3.org/1999/xhtml',
    'mathml': 'http://www.w3.org/1998/Math/MathML',
    'svg':  'http://www.w3.org/2000/svg'
     ...
  };
  return ns[prefix] || null;
}
```

PR #1678, opened 2016-09-04
refs #1677. Use the svg namespace resolver to find the svg node using xpath

sishen:
We use Document.evaluate to returns an XPathResult based on an XPath expression and other given parameters.

```js
var xpathResult = document.evaluate(
 xpathExpression, 
 contextNode, 
 namespaceResolver, 
 resultType, 
 result
);
```
**namespaceResolver** is a function that will be passed any namespace prefixes and should return a string representing the namespace URI associated with that prefix. It will be used to resolve prefixes within the XPath itself, so that they can be matched with the document. null is common for HTML documents or when no namespace prefixes are used.

So for svg nodes, we should use namespace ```http://www.w3.org/2000/svg```.

### DIFF
--- a/modules/clientutils.js
+++ b/modules/clientutils.js
@@ -55,6 +55,10 @@
             41, 42, 43, 44, 45, 46, 47, 48, 49, 50, 51, -1, -1, -1, -1, -1
         ];
         var SUPPORTED_SELECTOR_TYPES = ['css', 'xpath'];
+        var XPATH_NAMESPACE = {
+            svg: 'http://www.w3.org/2000/svg',
+            mathml: 'http://www.w3.org/1998/Math/MathML'
+        };
 
         // public members
         this.options = options || {};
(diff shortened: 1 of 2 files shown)


modules/clientutils.js
```diff
@@ -559,14 +559,26 @@
...
+         *
+         * @param String        prefix   The namespace prefix
+         * @return the resolve namespace or null
+         */
+        this.xpathNamespaceResolver = function xpathNamespaceResolver(prefix) {
+          if (prefix === 'svg') {
+            return 'http://www.w3.org/2000/svg';
+          }
+        };
+
```
sishen: @jefleponot Good point! Will make the change and test with other namespaces too. Thank you.

sishen: @jefleponot Made the changes, please review. Thanks!

jefleponot:
Tank you @sishen

Issue #1576, opened 2016-05-07
Suite test results display before tests are run


entendu:
Using the sample code [here](http://docs.casperjs.org/en/latest/testing.html#browser-tests), incorrect results of the test suite are displayed before the tests are run:

```
$ casperjs test casperjs/google.js 
Test file: casperjs/google.js                                                   
# Google search retrieves 10 or more results
PASS Google search retrieves 10 or more results (5 tests)
PASS google homepage title is the one expected
PASS main form is found
FAIL google title is ok
#    type: assertTitle
#    file: casperjs/google.js:12
#    code: test.assertTitle("casperjs - Recherche Google", "google title is ok");
#    subject: "Google"
#    expected: "casperjs - Recherche Google"
FAIL 3 tests executed in 0.502s, 2 passed, 1 failed, 0 dubious, 0 skipped.      

Details for the 1 failed test:

In casperjs/google.js:12
  Google search retrieves 10 or more results
    assertTitle: google title is ok

```

Notice that `PASS Google search retrieves 10 or more results (5 tests)` is printed before the tests are actually run.

When I update the sample code to use english & pass the tests, the same thing happens:

```
$ casperjs test casperjs/google.js 
Test file: casperjs/google.js                                                   
# Google search retrieves 10 or more results
PASS Google search retrieves 10 or more results (5 tests)
PASS google homepage title is the one expected
PASS main form is found
PASS google title is ok
PASS search term has been submitted
PASS google search for "casperjs" retrieves 10 or more results
PASS 5 tests executed in 1.093s, 5 passed, 0 failed, 0 dubious, 0 skipped.      
```

I noticed this happening on my own test suites and used this sample to verify.

This does NOT happen using the "[cow can moo](http://docs.casperjs.org/en/latest/testing.html#unit-testing)" example:
```
$ casperjs casperjs/cow-test.js
Test file: casperjs/cow-test.js                                                     
# Cow can moo
PASS Subject equals the expected value
PASS Subject is strictly true
PASS Cow can moo (2 tests)
PASS 2 tests executed in 0.026s, 2 passed, 0 failed, 0 dubious, 0 skipped.   
```

Platform: OSX 10.11.4
```
$ phantomjs --version
1.9.8

$ casperjs --version
1.1.0-beta4
```

morozgrafix:
Noticed that as well with `phantomjs 2.1.1` and `casperjs 1.1.1` on `OSX 10.11.4 (15E65`

[...]

PR #1693, opened 2016-10-01
Fix for "Suite test results display before tests are run" #1576

childen:
PASS test suite output is generated when test.done function is called. Before the message would be generated as soon as the callback function passed into test.begin returned which is too early async running tests.

### DIFF
--- a/modules/tester.js
+++ b/modules/tester.js
@@ -1128,15 +1128,6 @@ Tester.prototype.begin = function begin() {
         config = getConfig([].slice.call(arguments)),
         next = function() {
             config.test(this, this.casper);
-            if (!this.options.concise) {
-                this.casper.echo([
-                    this.colorize('PASS', 'INFO'),
-                    this.formatMessage(description),
-                    this.colorize(f('(%d test%s)',
-                                    config.planned,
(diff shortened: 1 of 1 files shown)


mickaelandrieu:
Hi @childen,

thank you for your contribution.

@jefleponot  what do you think ?

Mickaël

childen:
Do you mean "What functionality would we lose by getting rid of config.test"? The suggested change does not remove config.test. The only thing that is changed is that the "test suite passed output" is generated after test.done is called and not when the config.test function returns.

jefleponot:
nice cosmetic change
thanks for your contribution.

mickaelandrieu:
@jefleponot @BIGjuevos are you agree to merge this  one? :)

BIGjuevos:
> Do you mean "What functionality would we lose by getting rid of config.test"? The suggested change does not remove config.test. The only thing that is changed is that the "test suite passed output" is generated after test.done is called and not when the config.test function returns.

Thanks for explaining it! I wasn't extremely familiar with this bit of code and wanted to make sure we weren't accidentally overlooking something.

:+1: I am OK to merge this.

mickaelandrieu:
Thank you @childen and my mates for review :)

PR #1699, opened 2016-10-06
Fix some more links in the documentation

dritter:
- The Link to PhantomJS REPL was dead
- The Link to Homebrew led to an redirect


BIGjuevos:
Docs only, merging.

Thanks @dritter

Issue #1704, opened 2016-10-15
wrong signatures of event callbacks in docs


zzanzare:
see http://docs.casperjs.org/en/latest/modules/casper.html#onresourcereceived
onResourceReceived in fact receives a Casper object as the first argument, docs only talks about a resource object
onResourceRequested is missing a signature completely

Issue #1328, opened 2015-10-04
wrong signature for onError callback


nfriedly:
(I found a few different problems with the onError callback, so I'm a few tickets to make things easier to track.)

http://docs.casperjs.org/en/latest/modules/casper.html#onerror indicates that the `onError` callback is fired with the signature `onError(String msg, Array backtrace)` but the signature actually appears to be `onError(Object casper, String msg, String engine)` or something along those lines.

``` js
var casper = require('casper').create({
    onError: function(message, backtrace, thirdArg) {
        this.echo("onError - numArgs: " + arguments.length);
        this.echo("message: " + typeof message);
        this.echo(message);
        this.echo("backtace: " + typeof backtrace);
        this.echo(backtrace);
        this.echo("thirdArg: " + typeof thirdArg);
        this.echo(thirdArg);
    },
    waitTimeout: 100
});

casper.start('http://casperjs.org/', function() {
    //...
});

casper.waitForSelector('#id-that-doesnt-exist', function() {
    //...
});

casper.run();
```

outputs:

```
onError - numArgs: 3
message: object
[object Casper], currently at http://casperjs.org/
backtace: string
Wait timeout of 100ms expired, exiting.
thirdArg: string
phantom
```

Also, calling `JSON.stringify()` on the first Object throws another error, but that's not necessarily wrong behavior, not all objects can be stringified.

I'm on CasperJS v1.1.0-beta3.

zzanzare:
I believe this one can be closed as well now.

PR #1705, opened 2016-10-15
fixed signatures of event callbacks in docs

zzanzare:
refs #1704
refs #1328

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -110,7 +110,7 @@ Logging level (see the logging section for more information)
 
 **Default:** ``null``
 
-**Signature:** ``onAlert(String message)``
+**Signature:** ``onAlert(Object Casper, String message)``
 
 A function to be called when a javascript alert() is triggered
 
@@ -134,7 +134,7 @@ A function to be called when Casper#die() is called
 
 **Default:** ``null``
 
-**Signature:** ``onError(String msg, Array backtrace)``
+**Signature:** ``onError(Object Casper, String msg, Array backtrace)``
 
 A function to be called when an "error" level event occurs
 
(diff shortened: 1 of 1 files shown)


jefleponot:
??? it is not correct

zzanzare:
in which sense?

If it's about the travis failure, I believe my commit could not have possibly caused this error: `The command "npm install -g eslint@v1.10.1" failed and exited with 1` - can we somehow re-trigger the travis build?

jefleponot:
I'm sorry but you contribution has no sense
Javascript doesn't require as first parameter the caller so ... i don't understand your work.

Could you please add explanations ?

zzanzare:
well i was playing with casper, following the documentation here: http://docs.casperjs.org/en/latest/modules/casper.html#onresourcereceived and I needed to use these event handlers. And by dumping the function arguments I found out they are receiving the Casper object as the first argument, contrary to what the documentation said. So I fixed the documentation according to the actual behavior. See the tickets, they describe the problems quite clearly. It's just doc fix, no code change

jefleponot:
@zzanzare,

Thanks you for all yours explanations, I haven't understood that your are talking about casper.options.
You are totally right. Very goof job.

What do you think about the "onAlert" documentation ?
if you can fix it too as you would be nice :smile: 

Regards (before merge)

Could you please, after squash all commits and re-base for merging branch.

zzanzare:
@jefleponot ok I added also onAlert, and rebased

jefleponot:
@zzanzare , great job,

@mickaelandrieu @BIGjuevos ready to merge ?

mickaelandrieu:
👍 go for it!

Issue #1715, opened 2016-11-13
Local setUp/tearDown functions cannot be async


dritter:
If I specify a configuration object with async `setUp` and/or `tearDown` functions these never get called.

Example:
```JavaScript
casper.test.begin('setUp() tests', 1, {
  setUp: function(callback) {
      setTimeout(function() {
          setUp = true;
          callback();
      }, 50);
  },
  test: function(test) {
    test.assertTrue(setUp, 'Tester.setUp() executed the async setup function');
    test.done();
  }
});
```

In the [docs](http://docs.casperjs.org/en/latest/modules/tester.html#setup) there is no word about async setUp/tearDown working _only_ in the casperjs-global context.

kensoh:
Bumping up this interesting use case. @dritter I'm a CasperJS user and test automation engineer. Can you share more on your use case why async is needed for setup and teardown? I'll assume normally we want setup to be synchronously done before all the other testing happens.

kensoh:
Am thinking to close this, but other tools are also looking for have async setup/teardown. Looks like there are unique use cases for such and you've already committed much code. For example this https://github.com/nunit/nunit/issues/60. Also tagging current main maintainer @jefleponot.

PR #1718, opened 2016-11-14
Local config async

dritter:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | yes
| BC breaks?    | yes
| API-Change | yes
| Tests pass?   | yes
| Fixed tickets | #1715 

This PR cleans adds the general ability for all setUp/tearDown functions to be asynchronous.
I introduced a BC break here, because all setUp/tearDown function get the `tester` and `casper` instance passed, and as third argument the callback. This is done to clean up the code and to clarify the usage.
I know I could have avoided that BC break by checking if the first param is a function, and if so, assume that this is a callback, but the code is way more expressive now.

### DIFF
--- a/docs/modules/tester.rst
+++ b/docs/modules/tester.rst
@@ -679,6 +679,28 @@ A more asynchronous example::
         }
     });
 
+This could even be asynchronous functions:
+
+    var asyncCow;
+    casper.test.begin('Async Cows can fly', 2, {
+        setUp: function(test, globalCasper, callback) {
+            setTimeout(function() {
+                asyncCow = new AsyncCow();
+                callback();
+            }, 10);
+        },
+        tearDown: function(test, globalCasper, callback) {
+            setTimeout(function() {
+                asyncCow.destroy();
+                callback();
+            }, 10);
+        },
+        test: function(test) {
+            test.assertEquals(asyncCow.fly(), 'swoosh!');
+            test.done();
+        }
+    });
+
 .. index:: Colors
 
 ``colorize()``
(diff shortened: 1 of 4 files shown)


jefleponot:
very good job, but BC is a good thing
@mickaelandrieu @BIGjuevos what do you think ?

mickaelandrieu:
Hello @dritter, 

I don't have time yet to review this (sorry).

Is it a BC from a "code" point of view, or does it change the actual behavior?

I'm not really easy to bump 2.0 now, we don't have active maintainers (except @jefleponot).

Mickaël

BIGjuevos:
If we break BC, this would have to be the start of the 2.0 branch.

I really like the concept of this PR, and would like to see it make it in. If we could make it BC compat, I'd be onboard with this change.

dritter:
Hi all!

@jefleponot I agree, BC break is bad. But, as said, there is a way around it. But I am not a fan of lets check every argument and guess what we should do with it (like testing for function and execute it; what the BC way would look like).

@mickaelandrieu This PR indeed contains two changes regarding the usage:

First, for asynchronous setUp/tearDown, instead of calling 
```JavaScript
casper.test.setUp(function(done) { ... });
```
one must provide the callback function as third parameter. This is the obvious BC break.
```JavaScript
casper.test.setUp(function(test, globalCasper, done) { ... });
```

Secondly, as far as I can see, it is possible to define two `tearDown` functions. A local and a global one, which would be executed both. With this PR this is not possible any more. Either have a local or a global `tearDown` function.

So, there is no real change in behaviour.

kensoh:

PR #1738, opened 2016-12-20
Fix script errors processing

goganchic:
Use console.log instead of console.error for logging stacktrace.


mickaelandrieu:
Thank you @goganchic

Issue #1719, opened 2016-11-15
Should declare dependency on PhantomJS


zaz:
The [`phantomjs-prebuilt`](https://www.npmjs.com/package/phantomjs-prebuilt) package should be added as a dependency, as CasperJS doesn't seem to work at all without it.

If there's some practical reason for not doing this, it should at least be added as an optional dependency.

jefleponot:
Please Proposes a PR

zaz:
@jefleponot: Okay. PR #1743

jefleponot:
@mickaelandrieu , do you agree to add this feature ?

alexs20:
It is still has problem with phantomjs-prebuilt. I have phantomjs-prebuilt installed but not phantomjs as my karma-jasmine using only latest version.
After installing CasperJS and running "casperjs selftest" i got he flowing message:
"Fatal: The system cannot find the file specified; did you install phantomjs?"

jefleponot:
Hi @alexs20 

Try with environment variable : PHANTOMJS_EXECUTABLE=../phantomjs-prebuild/bin/phantomjs

But I'm not sure that "casperjs selftest" will success because phantomjs-2.5.0 isn't stable

kensoh:
Closing this for now as the dependency has already been merged and no other issues feedback, but do reopen if anything!

PR #1743, opened 2017-01-04
+ dependency: PhantomJS

zaz:
Issue #1719

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,9 @@
     "type": "git",
     "url": "git://github.com/casperjs/casperjs.git"
   },
+  "dependencies": {
+    "phantomjs-prebuilt": "2.1.x"
+  },
   "licenses": [
     {
       "type": "MIT",


jefleponot:
Hi @zaz ,

Could you explain me what is "phantomjs-prebuilt" and what is dependency ?
I don't know how does node module works ...

jefleponot:
Yes but CasperJS can also work with slimerJS , is there any dependencies for slimerJS ?

zaz:
@jefleponot: This PR doesn't add any dependencies for SlimerJS.

If that is the case, then this PR is inappropriate and an alternative method should be considered for ensuring the user installs PhantomJS or SlimerJS.  This could either be a message displayed to the user after install, or two new packages, _CasperJS_Phantom_ and _CasperJS_Slimer_ that wrap this package and the respective dependency.

jefleponot:
@zaz, thank for your explanations.
I hope that CasperJS could be also compatible with web extensions (Chrome and Firefox)

PR #1754, opened 2017-01-20
update env parameters to all scripts and executable

jefleponot:
Allows to users to add temporal environment variables with --env= parameter.

It is useful to test different proxies 
example : --env="htp_proxy='http://my.proxy:3128'" --env="no_proxy='127.0.0.1,.google.fr'

It is useful to test different routes in binding different local IP
example : --env='BIND_ADDR="X.X.X.X"' --env="LD_PRELOAD='/usr/lib/bind.so'"


jefleponot:
Add nodeJS script to launch casperJS and update python+exe+doc

PR #1755, opened 2017-01-21
Release 1.1.4

jefleponot:
preparing new release
changelog :
- real popups management
- fix bus : frame / die method ...
- add new events
- add parameters to launcher
...

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "casperjs",
   "description": "A navigation scripting & testing utility for PhantomJS and SlimerJS",
-  "version": "1.1.3",
+  "version": "1.1.4",
   "keywords": [
     "phantomjs",
     "slimerjs",


mickaelandrieu:
@bguerin it's not enough, we need to build or deprecate casperjs.exe first or Windows user will still stay in 1.1.2 because we already forgot them the last release.

Also, I wonder if @jefleponot have access to npm in order to declare the new release in node :)

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "casperjs",
   "description": "A navigation scripting & testing utility for PhantomJS and SlimerJS",
-  "version": "1.1.3",
+  "version": "1.1.4",
   "keywords": [
     "phantomjs",
     "slimerjs",
```
mickaelandrieu: you need to build a new version of casperjs.exe too

jefleponot:
@mickaelandrieu , you are wrong : bootstrap.js read the version number from package.json
Try to change the file, you will see ...

The real trouble is that we have add new parameters in the python script (for linux) and not in casper.exe (for windows)

Is anybody known how to build a portable executable ? Is there a documentation ?
Have you got an idea to create an universal launcher (without any compilation) ?

bguerin:
@mickaelandrieu yes I remember this problem of the exe wrapper having the bad version. So I did a grep 1.1 in the whole source tree and found nothing but the package.json. Make me curious so I looked at the code, and I think @jefleponot is right, this is exactly what I saw.
But I am always a little bit suspicious with Microsoft compiled code, resource files and all the dark magic behind each line the Microsoft compiler toolchain prints ...

bguerin:
However, at least a markdown file listing all that must be done for a release would be a good thing to have. Even better, a script which automate the stuff. I would love to help writing this :)

jefleponot:
Hi @bguerin, 
Do you known how to compile casperjs.exe 
mono / Visual C# ... ?

bguerin:
https://github.com/casperjs/casperjs/blob/master/Makefile#L17

mickaelandrieu:
If we compile the script using mono and test the .exe, I think it's enough for me.
Also, we need rights on npm to declare the new release, isn't it? (I don't use node, I'm an ashamed PHP developer :p )

jefleponot:
Hi @bguerin 
The makefile use "Mono" to test c# program but when I try to compile under windows 7 x86 : a nice error appears : **System.MissinfMethodException System.Array.Empty()**

Mono version 4.6.2
.NET 4

do you have any idea ?

bguerin:
No sorry ... I never code in C# :(

mickaelandrieu:
ping @rhwy would you mind to take a quick look? this is probably nothing but this will be great to keep a Windows exe :)

Thank you !

rhwy:
@mickaelandrieu I just tested compile+exec on your casperjs.cs file and i works, didn't have the same error messages. It's ugly but it works ;-)

but from what I saw on your csharp code, it's only a wrapper doing a shell call to something else, right? what's the purpose of that exe? why not using something like [enclosejs](http://enclosejs.com/) to build an exe if the point is to have an exe instead of maintaining this cs file that not everybody understands?

mickaelandrieu:
@rhwy sadly CasperJs is not a node module, it's only - as you say - a simple wrapper to make CasperJs available for Windows .

Thank you for enclosejs, this is probably what we need and this may works even if it's not a node module :)

mickaelandrieu:
@jefleponot I'll try to compile it in my own computer monday, let's keep this .exe file for this release and open an issue to remove it.

imho, we should allow only 1 way to install and use CasperJs : ``npm install -g casperjs`` and let Windows users deal with their "PATH" issues on their own.

jefleponot:
Hi @rhwy,

I add a new launcher file in nodejs (compatible with phantomjs) called "casperjs.js" in the PR #1754
But when I use enclosejs a message is displayed : "version evaluation please ..."
Is "enclosejs" free to use ? could we see code source ? is there any solution to compile js code for windows ?

rhwy:
Hi @jefleponot,
sorry for the misunderstanding, my idea with enclosejs was only to say that you should look in that direction instead of maintaining the c# wrapper, not to use it directly. As you said, it's a paid product and you should avoid that ;-)

jefleponot:
Hi @rhwy, Do you know an alternative tools to compile node js ?

PR #1756, opened 2017-01-23
Add setFilter('page.filePicker')

jefleponot:
When you click on a file input an event is triggered. This new filter allows to catch this event.

### DIFF
--- a/tests/suites/casper/formfill.js
+++ b/tests/suites/casper/formfill.js
@@ -1,6 +1,7 @@
 /*eslint strict:0*/
 var fs = require('fs');
 var selectXPath = require('casper').selectXPath;
+var exp = false;
 
 function testFormValues(test) {
     test.assertField('email', 'chuck@norris.com',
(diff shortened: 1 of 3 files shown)


tests/suites/casper/formfill.js
```diff
@@ -73,10 +84,15 @@ casper.test.begin('fill() & fillNames() tests', 18, function(test) {
             strange:       "very"
         });
         testFormValues(test);
-        if (!skipPhantom200(test, 1)) {
+        if (!skipPhantom200(test, 2)) {
             test.assertEvalEquals(function() {
                 return __utils__.findOne('input[name="file"]').files.length === 1;
             }, true, 'can select a file to upload');
+            if (!skipSlimer095(test,1)) {
```
mickaelandrieu: this if statement should be outside of skipPhantom200 check, isn't it?

jefleponot: @mickaelandrieu , I found the error... thanks.

jefleponot: Do you think it is normal that the "setFieldValue"  method doesn't work on input[type="file"] ?

mickaelandrieu: hmm... I guess it's a known issue on phantomjs /c @Vitallium

Vitallium: @mickaelandrieu this should be fixed in 2.5 version.

mickaelandrieu: Thank you :)

jefleponot:
@mickaelandrieu, do you agree to merge ?

mickaelandrieu:
thank you !

PR #1758, opened 2017-01-25
Fix area tag

jefleponot:
The "Aera" tag is always a hidden tag but we have to be able to click on it so ... a PR


mickaelandrieu:
Nice catch :)

PR #1788, opened 2017-03-16
Add cookbook example for passing in parameters into tests

### LINKED ISSUE #1787: How do I pass custom arguments into a testscript
Hi, is there a way for me to pass custom arguments to a casperjs test? 

I want to pass in the base url for the page I'm testing.

I tried using envitonment variables but process.env was not available. What is the suggested way?

tarjei:
Fixes #1787 :)

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -237,3 +237,4 @@ Let's say you want to be able to change the Uri your tests visits depending on w
     casper.test.begin(...)
     
 
+You can find the complete documentation for the cli object in http://docs.casperjs.org/en/latest/cli.html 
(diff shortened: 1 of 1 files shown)


jefleponot:
@tarjei 

What do you think of http://docs.casperjs.org/en/latest/cli.html ?

tarjei:
Looks good - but I didn't find that text when looking :) Maybe I should add a link in the cookbook text?

jefleponot:
a new PR is welcome :smile:

tarjei:
Updated.

tarjei:
Also - why isn't casper.once() documented?

jefleponot:
Casper is extended by EventEmitter (an old version) and it has never been documented ... :cry:

tarjei:
Ah :)

PR #1803, opened 2017-03-30
pageSettings DOCs update

hlr1983:
Should  `pageSettings` properties `javascriptCanOpenWindows` and `javascriptCanCloseWindows` be in the docs? `phantom.defaultPageSettings` has both set to true, but I haven't found any documentation about it, in CasperJS docs or in PhantomJS docs.

(SlimerJS has it documented in https://docs.slimerjs.org/current/configuration.html)


hlr1983:
I think it should not be merged, as SlimerJS docs says:

> Note that even if settings and phantom.defaultPageSettings are usable, only few settings are taken account by webpage objects: javascriptEnabled, loadImages and userAgent.

I've done a small test and window.open() works when `javascriptCanOpenWindows` is false. So, I'm closing this PR.

PR #1806, opened 2017-04-01
waitFor() always runs testFx (even if it times out before first test)

### LINKED ISSUE #1805: Error when waitFor timeout equals or less than options.retryTimeout?
If I do

```
casper.waitFor( function test() {
            return true;
        }, function success() {
            console.log('waitFor test is true');
        }, function fail() {
            console.log('waitFor test failed');
        },
1);
```
It shows _waitFor test failed_. So I did the following loop:

```
var customTimeout=100;
(function waitForLoop() {
    casper.waitFor( function test() {
            return true;
        }, function success() {
            console.log('waitFor test is true with customTimeout ' + customTimeout);
            customTimeout--
            waitForLoop()
        }, function fail() {
            console.log('waitFor failed with customTimeout ' + customTimeout);
            waitForLoop();
        },
    customTimeout);
})()
```

When `customTimeout` is _20_ it enters in an infinite loop. 

I've made the commit in PR #1806, changing the `waitFor()` function.

What do you think? Should change `waitFor()` or it's the expected behavior of `waitFor()` ?

hlr1983:
fix #1805

### DIFF
--- a/docs/modules/casper.rst
+++ b/docs/modules/casper.rst
@@ -1394,6 +1394,7 @@ Supported events are ``mouseup``, ``mousedown``, ``click``, ``dblclick``, ``mous
 and for phantomjs >= 1.9.8 ``mouseenter``, ``mouseleave`` and ``contextmenu``::
 
 .. warning::
+
 The list of supported events depends on the version of the engine in use.
 Older engines only provide partial support. For best support use recent builds of PhantomJS or SlimerJS."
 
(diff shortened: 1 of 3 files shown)


hlr1983:
Which do you think is better, 820d0413a30f87dc904a5573b6e3cb908c4d0de7 or bf1ad1e838f3bd786dac2c55b082d14db6f70a46 ?

Thanks

hlr1983:
@jefleponot 
It's ready, I think. :-)

Issue #1559, opened 2016-04-26
Casper doesn't pass firefox cli args to Slimer


NeonPaul:
Slimerjs allows certain [firefox-specific args](https://docs.slimerjs.org/current/configuration.html#command-line-options) to be passed on the command line. Namely these are:

> `-P name` - Use the indicated profile
> `-CreateProfile name` - Create a new profile
> `-profile path` - Use the given directory as a profile

Casper tries to pass these on but because the arg name and value are separated by a _space_ (I'm launching using NodeJS's `child_process.spawn`) rather than an _equals_ character (see [bin/casperjs line 116](https://github.com/casperjs/casperjs/blob/master/bin/casperjs#L116)), it fails.

The workaround solution I have found for the `-P` argument is only to use a profile named `test`, passing `-P` and `test` as two separate arguments (with an additional `test` argument to activate the testing library).

(NB: something to watch out for is that slimer expects the arg name and value to actually be separate arguments - e.g. `"slimerjs" "-P" "myProfile" "myScript.js"` - [issue filed here](https://github.com/laurentj/slimerjs/issues/463))

BIGjuevos:
So, there was a pull request #1507 that supposedly solved this issue.

What versions of SlimerJS and CasperJS are you using.  This is in version 1.1.0 and above.  Can you try that version and let us know how it goes?

NeonPaul:
@BIGjuevos We are using `casperjs@1.1.0` and `slimerjs@0.9.6`

hugsbrugs:
Yes it's really anoying not being able to launch slimerjs from casperjs with this Profile option : 
```
casperjs --debug=true --engine=slimerjs -P MyProfile my-script.js
```
launches the browser to let me choose my profile and then quit : 
```
(process:7337): GLib-CRITICAL **: g_slice_set_config: assertion 'sys_page_size == 0' failed
JavaScript warning: resource://slimerjs/slUtils.jsm, line 155: JavaScript 1.7's let blocks are deprecated
JavaScript error: jar:file:///usr/local/lib/node_modules/slimerjs/lib/slimer/xulrunner/omni.ja!/components/XULStore.js, line 66: NS_ERROR_FAILURE: Component returned failure code: 0x80004005 (NS_ERROR_FAILURE) [nsIProperties.get]

(process:7337): GLib-CRITICAL **: g_slice_set_config: assertion 'sys_page_size == 0' failed
[...]

PR #1818, opened 2017-04-19
Fix slimerjs arguments

chipironcin:
Allow CasperJS to detect slimerjs command line arguments with space between the argument and its value
Also improve logging by using Python logging utility instead of print (recommended practice) (found issue not flushing to stdout in some environments)

Should fix issue https://github.com/casperjs/casperjs/issues/1559

TODO:
- [ ] Apply same changes to casperjs.js and casperjs.exe

Advice requested for `casperjs.js` and `casperjs.exe`

### DIFF
--- a/bin/casperjs
+++ b/bin/casperjs
@@ -98,7 +98,13 @@ SUPPORTED_ENGINES = {
             'w',
         ],
         'env_varname': 'SLIMERJS_EXECUTABLE',
-        'default_exec' : 'slimerjs'
+        'default_exec' : 'slimerjs',
+        'native_args_with_space': [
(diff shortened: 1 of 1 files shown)


jefleponot:
@chipironcin 
Very Good job

chipironcin:
What about 
>Advice requested for casperjs.js and casperjs.exe
Should we also apply these changes to those files?

jefleponot:
@chipironcin 

It is done for casperjs.js :+1: 
and for casperjs.exe : Do you code in mono Csharp ?

chipironcin:
nope 😂

PR #1837, opened 2017-06-03
Fixes #1222 - Improve visibility tests - correctly report if descendent of hidden el…

### LINKED ISSUE #1222: clientutils.elementVisible() gives false positive when container has display:none
I was putting some tests together, and noticed that an element was being reported as visible, even though it was not.  This happened when I checked an element (`<a>`) within a container that was hidden (`display:none`).

When the container's visibility is `hidden`, then `getComputedStyle` works.

We should also update the test cases in visible.js.  `waitWhileVisible is misleading because the element starts as hidden, and we're not testing that any waiting happened.  We could also add some more test cases, such as testing an element that is a child of a hidden container, zero width/height, opacity, and possibly others.

I began working on the code, but haven't had enough time to finish it, and it might be a while (maybe a week or two) before I get another opportunity.

Rykus0:
…ement as not visible, as well as zero-width/height or empty-content elements

### DIFF
--- a/modules/clientutils.js
+++ b/modules/clientutils.js
@@ -155,7 +155,7 @@
          */
         this.elementVisible = function elementVisible(elem) {
             var style;
-            try {
+            try { 
                 style = window.getComputedStyle(elem, null);
             } catch (e) {
                 return false;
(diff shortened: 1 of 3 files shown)


Rykus0:
I just didn't realize the original had changed from when I first started this.  I'm going to close this one and submit another.

Issue #1844, opened 2017-07-01
Possible EventEmitter memory leak detected


kensoh:
Posting this issue on behalf of David @djwales who raised this in CasperJS mailing list. He's open to sending PRs for this and I thought here would be better for maintainers @jefleponot to have a look.

---

Hi,

I have been using CasperJS with PhantomJS for a number of weeks and everything was working as expected. However, I encountered a problem yesterday that took a lot of investigating before I could resolve the problem. I started receiving the following message after some more tests were added (added to the same file as other tests):

```
(node) warning: possible EventEmitter memory leak detected. %d listeners added. Use emitter.setMaxListeners() to increase limit.
```
I spent a lot of time googling for information on the issue but kept getting results for the equivalent problem in node based applications. I could not find anything for CasperJS/PhantomJS. I eventually decided to look into the CasperJS source and managed to find where the message was being raised. It was coming from the events.js file in modules directory. After a bit more searching around the source I discovered that I could increase the max listeners by calling the setMaxListeners method on the casper prototype. This then resolved my problem. 

Just for reference, I pulled CasperJS and PhantomJS into my project from NPM and invoke casper from an NPM script. The mention of node in the console output along with this fact added more confusion because I wasn't sure if it was something to do with node (via NPM) or CasperJS/PhantomJS.

The point I would like to raise in here is that though the error message was correct, the suggestion was misleading and didn't help me to resolve the problem. This is how the error message presented itself in my console output:

```
(node) warning: possible EventEmitter memory leak detected. %d listeners added. Use emitter.setMaxListeners() to increase limit. 11
  <rest of path>/src/client/phantomjs:/platform/events.js:184 in addListener
FAIL (node) warning: possible EventEmitter memory leak detected. %d listeners added. Use emitter.setMaxListeners() to increase limit. 11
Failed on url: http://localhost:8080/criminal/case-details
#    type: error
#    file: <rest of path>/src/client/test/e2e/dist/scenarios/test.scn.js
#    subject: false
#    error: "(node) warning: possible EventEmitter memory leak detected. %d listeners added. Use emitter.setMaxListeners() to increase limit. 11"
#    stack: in addListener() in phantomjs://platform/events.js:184
```

It tells the user to use emitter.setMaxListeners, but I was not able to determine how to do this. All attempts to search through google just showed how to do it in a node based application. The line starting with "stack" also confused me because I couldn't determine where the file was located, though this is probably down tomy lack of experience with PhantomJS.

[...]

PR #1848, opened 2017-07-07
Possible EventEmitter memory leak - Refs #1844

djwales:
PR for issue #1844 

Changes as per issue. Modified the warning messages in modules/events.js when max listeners is exceeded. Also updated the online documentation to discuss the setMaxListeners function.


djwales:
@jefleponot 

Travis has randomly failed again on the SlimerJS tests. What should I do now?

PR #1857, opened 2017-07-24
Is phantomJS 2.5 buildworks on CasperJS ?

jefleponot:
https://github.com/nkovacs/selenium-standalone-phantomjs/commit/11420aaf2174428b56086673232365c820f98a42

https://github.com/ariya/phantomjs/issues/14459#issuecomment-286499219


jefleponot:
@kensoh 

Do you agree with this PR ?

kensoh:
Thanks @jefleponot! The PR looks like enabling CasperJS support for PhantomJS v2.5 engine built by @nkovacs plus some improvements not related to v2.5. Will enabling cause more issues for users who use it because v2.5 has lesser compatibility score than v2.1.1 from above link? (https://github.com/nkovacs/selenium-standalone-phantomjs/commit/11420aaf2174428b56086673232365c820f98a42)

I'm not technically familiar with CasperJS code base to comment, maybe CasperJS previous maintainers can have a look to comment. Below are some notes from my limited knowledge of CasperJS's internals - 

- there are a few references to `utils.gteVersion` which has a small typo instead of `utils.getVersion`. also, this method does not exist currently in utils.js.
- there are some special handling for v2.2.0, and v2.x.x, I assume they are meant to address special situations for both and combined into the PR.
- there is a reference to a webpage on localhost server in multiple-popups.html but not in any other files in that folder. also, in suites only http_status.js creates a local webserver. not sure if that test will work or if it already takes care of creating a local webserver to serve that page.

jefleponot:
@kensoh 

When I try to integrate PhantomJS 2.5.0, I found bugs with PhantomJS 2.X.X so that explains second point of your comment.
multiple-popups.html occurs trouble usually because of opened port so I searched a way to fix this trouble.

kensoh:
Thanks @jefleponot I guess users who choose to use PhantomJS v2.5 have to know that it has many compatibility issues and not really in a usable state. They might still want to use v2.5 engine for some specific features but the pros / cons have to be considered by them before using.

jefleponot:
@mickaelandrieu @BIGjuevos @n1k0 @kensoh @CodersBrothers @RichardJECooke @mlalpho

Is somebody interested in using Promises with CasperJS thougth PhantomJS 2.5.0 ?

Please Help me in reviewing ...

jefleponot:
@kensoh 

I tested PhantomJS 2.5 standalone and I found some bugs that CasperJS has easily workaround.
it is now fully usable with CasperJS.

Do you think you could test it with yours scripts ?

kensoh:
I have 8-9 sample scripts that test a number of CasperJS functionalities, but only a small subset of all CasperJS methods. I tried running PhantomJS v2.5 from https://github.com/casperjs/phantomjs-builds/releases/tag/v2.5.0 and CasperJS from https://github.com/jefleponot/casperjs/tree/phantom2.5 but get below error on my CentOS 6 Linux server.

`error while loading shared libraries: libjpeg.so.8: cannot open shared object file: No such file or directory`

I know that this PhantomJS build is for Linux and there isn't one for macOS or Windows. The error looks like the build is somehow not self-contained / compatible with at least the Linux version I'm using (CentOS 6).

jefleponot:
@kensoh 

You are right. The standalone linux version just works with ubuntu 16.XX

mickaelandrieu:
@jefleponot sounds interesting, what are the improvements of Phantomjs 2.5 vs Phantomjs 2.2.

Also, phantomjs is quite abandonned right now: I think we may(should?) use slimerjs as main engine for the next releases of CasperJS, what do you think?

jefleponot:
Hi @mickaelandrieu,
It's been a long time that you have post issue. 
You always works for Prestashop ?

These PR just fix some PhantomJS 2.5 bugs (fileFieldSelector ...)

SlimerJS is really nice now because it can work in normal mode or headless mode.
I had changed job last september, so I haven't no more time to animate and to improve CasperJS.
I will definitely move project as not maintained one.

mickaelandrieu:
> I will definitely move project as not maintained one.

Yes, I think it's time: thanks for all the work you have done on this project.

Before making this project unmaintained, you should add yourself to the list of maintainers in the README.

For me, 👍 on making this project unmaintained: there are better solutions for everything we can do with CasperJS right now, so let's move on :)

kensoh:
Hi @jefleponot and @mickaelandrieu for me I haven't found something better than PhantomJS + CasperJS framework. The beauty is you can simply unzip and run, without having other developer environments and dependencies such as Node.js or other language frameworks. This alone is a huge win in many organizations where users do not have admin access to their laptops to install software.

That is why I continue to build a free RPA tool on top of this stack - https://github.com/kelaberetiv/TagUI

The tool now is being maintained by my non-profit organization funded by Singapore government. I'll hope to have CasperJS and PhantomJS having compiled versions for operating systems in future. And if that sort of effort requires payment, it is really worthwhile to pay for such newer compiled versions.

kensoh:
Adding on, the AI representative body for Singapore government (AI Singapore) will be meeting the AI representative body for France soon. There is likely some agreement of collaboration between our countries. I'll be sure to tell my bosses of the good work that @n1k0 and you guys did, and hopefully keep the project alive. PhantomJS and CasperJS were created in an era before the tech giants took over the open-source world, and even today you see new stars and users every day for CasperJS. That shows that the software foundation, design, and documentation is still relevant today.

PR #1873, opened 2017-08-23
update cli.rst to properly format sample code block

kensoh:
I've made a PR which somehow doesn't respect the Sphinx documentation syntax. Making an update so that the sample code block shows up as code block instead of plain text. I'm sorry for the oversight! It is my first time updating Sphinx documentation and I was not able to preview the actual change live. https://github.com/casperjs/casperjs/pull/1871/files

### DIFF
--- a/docs/cli.rst
+++ b/docs/cli.rst
@@ -148,7 +148,7 @@ If you run this script:
 
 As you can see, the ``01234567`` value has been cast to a *Number*.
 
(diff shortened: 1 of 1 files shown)


kensoh:
Hi @jefleponot I've made an edit to sub-branch but somehow the tests fail. Although the original edit pass. I wasn't able to edit on my fork as it doesn't show up correctly when doing diff.

kensoh:
Committing now as the tests now pass.

PR #1910, opened 2018-01-29
Slimer js 1.0.0

jefleponot:
First test before headless parameter


jefleponot:
Add The new version 1.0.0 of slimerJS

Issue #1912, opened 2018-02-06
Missing dynamic content in downloaded HTML


Zearin:
I’m scraping content from a Q&A-style site, where the site reorders the answers for any question dynamically.  

When I take a screenshot of the page to verify that the content I want is visible, it shows up as expected.  Everything appears exactly as if I open the page in the browser.

However, when I use `casper.download()`, the page does _not_ include the answer I am looking for.  It seems to download the page as it would appear if I load the page in a browser with javascript disabled.  The problem was unchanged after I turned off the “web security” option.

Am I missing something?

kensoh:
Hi @Zearin, the download() function is normally used to download a static resource from what I understand. Why not try using casper.getHTML() instead if you want to get the HTML DOM?

Zearin:
…Oh. ☹ After all the stuff I read about using Casper to modify the remote DOM , I thought `download()` would be a part of that.  

So, it just sends a _separate_ “dumb” request?

That’s okay, I guess.  But I could just use `curl` if I wanted that.

I _could_ use `casper.getHTML()`, but the reason I specifically wanted to use `download()` was because it let me specify where to save the file at the same time.  

---- 

I spent a _lot_ of time trying to figure this out before opening this issue.  I kept thinking I must be doing something wrong, or I missed something, or maybe there was some undocumented function…or _something_.

It might help other people from going through the same trouble I did if the documentation for `download()` clearly states that it doesn’t do anything to “living DOM” in Casper’s headless browser, but is just for sending a simple request and saving the result to disk.  

In fact, let’s make it official…

## Feature Request

- [ ] Clearly document `casper.download()` as a separate request that immediately saves the response to disk, and does not interact with page instances (and therefore cannot saved a DOM-manipulated page).

kensoh:
Hi @jefleponot I've updated the documentation for download() with this note - `Also, this is meant more for retrieving static resources. If you want to get the JavaScript-rendered HTML DOM use getHTML() instead.`

http://docs.casperjs.org/en/latest/modules/casper.html#download

Zearin:
Thank you @kensoh :D

kensoh:
Thanks @Zearin for flagging this out! :smile:

PR #1914, opened 2018-02-10
#1912 - documentation improvement for download()

kensoh:


kensoh:
Hi @jefleponot this PR is only documentation update, not sure why master branch is failing. Copying you to see if there is anything I can follow-up with.

